PR #1, opened 2021-01-29
Cython bindings for libbson

prashantmital:
PR that adds rudimentary bindings for libbson and also an example of how to use them from a C++ cython project.

In order to setup the project you will need to download libbson builds from https://evergreen.mongodb.com/waterfall/mongo-c-driver-r1.17:

- Start by identifying a `debug-compile` task build for the platform you are looking to use. E.g. for macOS you can use [this build](https://evergreen.mongodb.com/task/mongo_c_driver_r1.17_darwin_debug_compile_sasl_darwinssl_d53b3dc78a3dc56bcafbe5b768b3dfe62985d67a_20_12_02_13_35_28)
- Download the attached `tar.gz` file and unzip it. 
- Note the abspath to the unzipped tree - e.g `/path/to/cdriver`

A simpler way to do this on macOS is to use homebrew:

```
$ brew install mongo-c-driver
```

To run the included test:

- Navigate to the `mongo-arrow/bindings/python`
- Run `CFLAGS=-I/path/to/cdriver/install-dir/include/libbson-1.0 python setup.py build_ext --inplace --force` 
- If you used Homebrew to install libbson use the following command: `CFLAGS=-I/usr/local/include/libbson-1.0 python setup.py build_ext --inplace --force` 
- Run `LDFLAGS=-L/path/to/cdriver/install-dir/lib python -m unittest discover test`
- If you used Homebrew to install libbson use the following command: `python -m unittest discover test`


prashantmital:
Note that unlike cffi and ctypes, Cython does not allow setting the path at which to find the shared library object to be imported at runtime. Consequently, I have included the `extra_link_args` hack in this PR. In order to actually distribute the wheels, we will need to incorporate the [delocate utility](https://github.com/matthew-brett/delocate) on macOS and [auditwheels](https://github.com/pypa/auditwheel) repair functionality into the build process to 'fix' the path from which to load libbson at runtime.

bindings/python/setup.py
```diff
@@ -0,0 +1,35 @@
+from setuptools import find_packages, setup, Extension
```
ShaneHarvey, line 1: Please add our standard LICENSE file and add the license header to all files.

bindings/python/setup.py
```diff
@@ -0,0 +1,35 @@
...
+from Cython.Build import cythonize
+
+import os
+
+
+LIBBSON_PATH = os.path.join(
+    os.path.dirname(os.path.abspath(__file__)), 'pymongoarrow/libbson/lib')
```
ShaneHarvey, line 8: Does this have to be an absolute path? If we make this a relative path (and include the libbson file in the wheel) will cython work correctly?

bindings/python/setup.py
```diff
@@ -0,0 +1,35 @@
...
+              libraries=['bson-1.0'],
+              library_dirs=[LIBBSON_PATH],
+              extra_link_args=['-Wl,-rpath,' + LIBBSON_PATH])
+]
+
+# ext_modules = cythonize("pymongoarrow/*.pyx")
+# for module in ext_modules:
+#     module.include_dirs.append('pymongoarrow/libbson/include')
+#     module.libraries.append('bson-1.0')
+#     module.library_dirs.append('pymongoarrow/libbson/lib')
```
ShaneHarvey, line 25: Is this needed?

bindings/python/setup.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+# import ipdb; ipdb.set_trace()
```
ShaneHarvey, line 28: Please remove.

bindings/python/pymongoarrow/bson_util.pyx
```diff
@@ -0,0 +1,33 @@
...
+
+    cdef cbool reached_eof = False
+    cdef const bson_t *doc = bson_reader_read(stream_reader, &reached_eof)
+    cdef bson_iter_t doc_iter
+
+    documents = []
+
+    while doc != NULL:
+        if not bson_iter_init(&doc_iter, doc):
+            raise RuntimeError
```
ShaneHarvey, line 20: Please add a helpful error message to this RuntimeError.

bindings/python/pymongoarrow/bson_util.pyx
```diff
@@ -0,0 +1,33 @@
+# distutils: language=c++
+# cython: language_level=3
+
+from libcpp cimport bool as cbool
+
+from pymongoarrow.libbson cimport *
+
+
+cdef list _docs_from_bson_stream(const uint8_t *docstream, size_t length):
+    cdef bson_reader_t *stream_reader = bson_reader_new_from_data(docstream, length)
```
ShaneHarvey, line 10: Does this need a NULL check?

ShaneHarvey (commented):
Neat!

bindings/python/setup.py
prashantmital: It seems this will not be sufficient. The correct way to distribute shared library objects is using delocate or audit-wheel as I have mentioned in my comment.

bindings/python/setup.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+EXT_MODULES = [
+    Extension('pymongoarrow.bson_util',
+              sources=['pymongoarrow/bson_util.pyx'],
+              language='c++',
+              include_dirs=['pymongoarrow/libbson/include'],
+              libraries=['bson-1.0'],
+              library_dirs=[LIBBSON_PATH],
+              extra_link_args=['-Wl,-rpath,' + LIBBSON_PATH])
```
ShaneHarvey, line 18: Are library_dirs and extra_link_args needed at all here? The docs suggest that only `libraries` is required for cython to find the library: https://cython.readthedocs.io/en/stable/src/tutorial/clibraries.html#dynamic-linking

> Copy libbson-1.0.0.<extension> to mongo-arrow/bindings/python/pymongoarrow/libbson/lib

I think if we remove library_dirs and extra_link_args then we can remove this requirement to copy libbson into the mongo-arrow dir.

Again from the cython dynamic-linking docs:

> If the libcalg is not installed in a ‘normal’ location, users can provide the required parameters externally by passing appropriate C compiler flags, such as:
  ```
  CFLAGS="-I/usr/local/otherdir/calg/include"  \
LDFLAGS="-L/usr/local/otherdir/calg/lib"     \
    python setup.py build_ext -i
  ```
  
So if you prefer not to install libbson on your system you could do:
  ```
  CFLAGS="-I/usr/local/otherdir/libbson/include"  \
LDFLAGS="-L/usr/local/otherdir/libbson/lib"     \
    python setup.py build_ext -i
  ```

prashantmital: Thanks for this suggestion. This is how things are setup now. I have updated the PR description to reflect this change.

bindings/python/pymongoarrow/bson_util.pyx
prashantmital: `bson_util` has been removed since it wasn't intended to be a part of the final package and only served as an example.

bindings/python/pymongoarrow/bson_util.pyx
prashantmital: bson_util has been removed.

bindings/python/setup.py
prashantmital: Done.

bindings/python/setup.py
prashantmital: Cleaned up setup.py

bindings/python/setup.py
prashantmital: LICENSE has been added at the root of the project. Should I add another under the Python bindings? 

Added standard license header to all files.

bindings/python/.gitignore
```diff
@@ -31,8 +31,5 @@ MANIFEST
 *.c
 *.cpp
 
-# PyCharm
-*.idea/
-
```
ShaneHarvey, line 36: It would be good to keep this.

bindings/python/pymongoarrow/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+    from distutils.version import LooseVersion as _LooseVersion
+
+    def _parse_version(version):
+        return _LooseVersion(version)
+
+
+if _parse_version(libbson_version) < _parse_version(_MIN_LIBBSON_VERSION):
+    raise RuntimeError(
+        "Expected libbson version {} or greater, found {}}".format(
+            _MIN_LIBBSON_VERSION, libbson_version))
```
ShaneHarvey, line 31: I think using the library at import time like this can cause some problems down the line. Imagine when we want to integrate this library with pymongo. Then pymongo will look like this:

```
try:
    import pymongoarrow
except ImportError:
    pass  # not installed 
```

With the current design we need to catch RuntimeError as well which is a bit odd:
```
try:
    import pymongoarrow
except (ImportError, RuntimeError):
    pass  # not installed 
```

I suggest we defer calling into the library until it's first used. For example:
```
__version__ = bson_get_version().decode('utf-8')
```
Would become:
```
def libbson_version():
    return bson_get_version().decode('utf-8')
```

Feel free to open a follow up ticket if you like to merge this as is and discuss this issue further in Jira.

Edit: a simple alternative option might be to change this line to raise an ImportError instead of a RuntimeError.

ShaneHarvey (changes requested):

bindings/python/setup.py
ShaneHarvey: Nice!

bindings/python/pymongoarrow/bson_util.pyx
ShaneHarvey: Ack.

bindings/python/setup.py
ShaneHarvey: Yes the LICENSE needs to be picked up and distributed with the distribution, eg:https://github.com/mongodb/mongo-python-driver/blob/master/MANIFEST.in#L2

bindings/python/pymongoarrow/__init__.py
prashantmital: Good point except there is no one point of entry for loading libbson in this case so even if the version check is done lazily, it would have to be explicitly added to every possible method in the libbson bindings that could be directly called by external code.

I have changed it to raise an ImportError instead.

bindings/python/.gitignore
prashantmital: Reverted.

bindings/python/setup.py
prashantmital: Done.

ShaneHarvey (approved):

PR #3, opened 2021-02-10
POC implementation for converting a stream of BSON documents to a pyarrow.Table instance

prashantmital:
With this PR we can now move data from a stream of BSON documents, into Arrow arrays which can then be assembled as an Arrow Table. 

This current design is inefficient because:
1. Extensive use of Python maps. Looking up entities stored in `Schema` and `PyMongoArrowContext` objecs incurs the additional overhead of calling Python functions from C. We do this every time we need to process a field from the BSON stream so it will result in very high overhead for a large number of documents/fields. Mitigation: use `std::map` instances. instead of Python dicts.
2. Use of Python APIs for Builder classes. When appending entries via the builder APIs we are using the Cython wrapper classes around Arrow's native ArrayBuilder classes. In addition to overhead of the Python function call itself, this causes the values being appended themselves to be wrapped and unwrapped during each called to '_BaseBuilder.appendI()`. Mitigation: unwrap the native builder classes from the encapsulating Python objects (using `_BuilderBase.unwrap` methods) and operate on the native classes directly.

### DIFF
--- a/bindings/python/pymongoarrow/bson.pyi
+++ b/bindings/python/pymongoarrow/bson.pyi
@@ -13,12 +13,19 @@
 # limitations under the License.
 
 
+cdef const bson_t* bson_reader_read_safe(bson_reader_t* stream_reader, cbool* reached_eof) except *:
+    cdef const bson_t* doc = bson_reader_read(stream_reader, reached_eof)
+    if doc == NULL and dereference(reached_eof) == False:
+        raise RuntimeError("Could not read BSON stream")
+    return doc
+
+
 def process_bson_stream(bson_stream, context):
     cdef const uint8_t* docstream = <const uint8_t *>bson_stream
     cdef size_t length = <size_t>PyBytes_Size(bson_stream)
     cdef bson_reader_t* stream_reader = bson_reader_new_from_data(docstream, length)
     cdef cbool reached_eof = False
-    cdef const bson_t * doc = bson_reader_read(stream_reader, & reached_eof)
+    cdef const bson_t * doc = bson_reader_read_safe(stream_reader, &reached_eof)
     cdef bson_iter_t doc_iter
     cdef const char* key
     cdef bson_type_t value_t
@@ -69,6 +76,6 @@ def process_bson_stream(bson_stream, context):
             for _, builder in builder_map.items():
                 if len(builder) != count:
                     builder.append_null()
-            doc = bson_reader_read(stream_reader, &reached_eof)
+            doc = bson_reader_read_safe(stream_reader, &reached_eof)
     finally:
         bson_reader_destroy(stream_reader)
--- a/bindings/python/pymongoarrow/bson.pyi
+++ b/bindings/python/pymongoarrow/bson.pyi
@@ -64,3 +64,5 @@ def process_bson_stream(bson_stream, context):
                 else:
                     raise TypeError('unknown ftype {}'.format(ftype))
         doc = bson_reader_read(stream_reader, &reached_eof)
+
+    bson_reader_destroy(stream_reader)
--- a/bindings/python/pymongoarrow/bson.pyi
+++ b/bindings/python/pymongoarrow/bson.pyi
@@ -13,9 +13,10 @@
 # limitations under the License.
 
 
-cdef const bson_t* bson_reader_read_safe(bson_reader_t* stream_reader, cbool* reached_eof) except? NULL:
-    cdef const bson_t* doc = bson_reader_read(stream_reader, reached_eof)
-    if doc == NULL and dereference(reached_eof) == False:
+cdef const bson_t* bson_reader_read_safe(bson_reader_t* stream_reader) except? NULL:
+    cdef cbool reached_eof = False
+    cdef const bson_t* doc = bson_reader_read(stream_reader, &reached_eof)
+    if doc == NULL and reached_eof == False:
         raise RuntimeError("Could not read BSON stream")
     return doc
 
@@ -24,8 +25,7 @@ def process_bson_stream(bson_stream, context):
     cdef const uint8_t* docstream = <const uint8_t *>bson_stream
     cdef size_t length = <size_t>PyBytes_Size(bson_stream)
     cdef bson_reader_t* stream_reader = bson_reader_new_from_data(docstream, length)
-    cdef cbool reached_eof = False
-    cdef const bson_t * doc = bson_reader_read_safe(stream_reader, &reached_eof)
+    cdef const bson_t * doc = NULL
     cdef bson_iter_t doc_iter
     cdef const char* key
     cdef bson_type_t value_t
@@ -35,7 +35,10 @@ def process_bson_stream(bson_stream, context):
     type_map = context.type_map
 
     try:
-        while doc != NULL:
+        while True:
+            doc = bson_reader_read_safe(stream_reader)
+            if doc == NULL:
+                break
             if not bson_iter_init(&doc_iter, doc):
                 raise RuntimeError
             while bson_iter_next(&doc_iter):
@@ -75,7 +78,7 @@ def process_bson_stream(bson_stream, context):
             count += 1
             for _, builder in builder_map.items():
                 if len(builder) != count:
+                    # Append null to account for any missing field(s)
                     builder.append_null()
-            doc = bson_reader_read_safe(stream_reader, &reached_eof)
     finally:
         bson_reader_destroy(stream_reader)
--- a/bindings/python/pymongoarrow/bson.pyi
+++ b/bindings/python/pymongoarrow/bson.pyi
@@ -22,6 +22,7 @@ def process_bson_stream(bson_stream, context):
     cdef bson_iter_t doc_iter
     cdef const char* key
     cdef bson_type_t value_t
+    cdef uint64_t count = 0
 
     builder_map = context.builder_map
     type_map = context.type_map
@@ -64,6 +65,10 @@ def process_bson_stream(bson_stream, context):
                             builder.append_null()
                     else:
                         raise TypeError('unknown ftype {}'.format(ftype))
+            count += 1
+            for _, builder in builder_map.items():
+                if len(builder) != count:
+                    builder.append_null()
             doc = bson_reader_read(stream_reader, &reached_eof)
     finally:
         bson_reader_destroy(stream_reader)
--- a/bindings/python/pymongoarrow/bson.pyi
+++ b/bindings/python/pymongoarrow/bson.pyi
@@ -26,43 +26,44 @@ def process_bson_stream(bson_stream, context):
     builder_map = context.builder_map
     type_map = context.type_map
 
-    while doc != NULL:
-        if not bson_iter_init(&doc_iter, doc):
-            raise RuntimeError
-        while bson_iter_next(&doc_iter):
-            key = bson_iter_key(&doc_iter)
-            if key in builder_map:
-                builder = builder_map[key]
-                ftype = type_map[key]
-                value_t = bson_iter_type(&doc_iter)
-                if ftype == _BsonArrowTypes.int32:
-                    if value_t == BSON_TYPE_INT32:
-                        builder.append(bson_iter_int32(&doc_iter))
+    try:
+        while doc != NULL:
+            if not bson_iter_init(&doc_iter, doc):
+                raise RuntimeError
+            while bson_iter_next(&doc_iter):
+                key = bson_iter_key(&doc_iter)
+                if key in builder_map:
+                    builder = builder_map[key]
+                    ftype = type_map[key]
+                    value_t = bson_iter_type(&doc_iter)
+                    if ftype == _BsonArrowTypes.int32:
+                        if value_t == BSON_TYPE_INT32:
+                            builder.append(bson_iter_int32(&doc_iter))
+                        else:
+                            builder.append_null()
+                    elif ftype == _BsonArrowTypes.int64:
+                        if (value_t == BSON_TYPE_INT64 or
+                                value_t == BSON_TYPE_BOOL or
+                                value_t == BSON_TYPE_DOUBLE or
+                                value_t == BSON_TYPE_INT32):
+                            builder.append(bson_iter_as_int64(&doc_iter))
+                        else:
+                            builder.append_null()
+                    elif ftype == _BsonArrowTypes.double:
+                        if (value_t == BSON_TYPE_DOUBLE or
+                                value_t == BSON_TYPE_BOOL or
+                                value_t == BSON_TYPE_INT32 or
+                                value_t == BSON_TYPE_INT64):
+                            builder.append(bson_iter_as_double(&doc_iter))
+                        else:
+                            builder.append_null()
+                    elif ftype == _BsonArrowTypes.datetime:
+                        if value_t == BSON_TYPE_DATE_TIME:
+                            builder.append(bson_iter_date_time(&doc_iter))
+                        else:
+                            builder.append_null()
                     else:
-                        builder.append_null()
-                elif ftype == _BsonArrowTypes.int64:
-                    if (value_t == BSON_TYPE_INT64 or
-                            value_t == BSON_TYPE_BOOL or
-                            value_t == BSON_TYPE_DOUBLE or
-                            value_t == BSON_TYPE_INT32):
-                        builder.append(bson_iter_as_int64(&doc_iter))
-                    else:
-                        builder.append_null()
-                elif ftype == _BsonArrowTypes.double:
-                    if (value_t == BSON_TYPE_DOUBLE or
-                            value_t == BSON_TYPE_BOOL or
-                            value_t == BSON_TYPE_INT32 or
-                            value_t == BSON_TYPE_INT64):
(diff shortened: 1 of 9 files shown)


bindings/python/pymongoarrow/schema.py
```diff
@@ -0,0 +1,56 @@
...
+
+    @staticmethod
+    def _normalize_mapping(mapping):
+        normed = {}
+        for fname, ftype in mapping.items():
+            normalized_type = _normalize_typeid(ftype)
+            if normalized_type is None:
+                raise ValueError(
+                    "Unsupported type identifier {} for field {}".format(
+                        ftype, fname))
```
ShaneHarvey, line 37: Suggest moving the ValueError raising to `_normalize_typeid` itself to avoid duplication.

bindings/python/pymongoarrow/types.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class _BsonArrowTypes(enum.Enum):
+    datetime = 1
+    double = 2
+    int32 = 3
+    int64 = 4
+
+
+_SUPPORTED_TYPE_IDENTIFIERS = (Int64, float, int, datetime)
```
ShaneHarvey, line 31: Is this needed? It seems like we can use _TYPE_NORMALIZER_FACTORY instead, eg:
```
typeid in _TYPE_NORMALIZER_FACTORY
```

bindings/python/pymongoarrow/schema.py
```diff
@@ -0,0 +1,56 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import collections.abc as abc
+
+from pymongoarrow.types import _normalize_typeid
+
+
+class Schema:
```
ShaneHarvey, line 19: Can you add some tests for this class? It's a little hard to understand what schemas are supported.

bindings/python/pymongoarrow/bson.pyi
```diff
@@ -0,0 +1,66 @@
...
+        if not bson_iter_init(&doc_iter, doc):
+            raise RuntimeError
+        while bson_iter_next(&doc_iter):
+            key = bson_iter_key(&doc_iter)
+            if key in builder_map:
+                builder = builder_map[key]
+                ftype = type_map[key]
+                value_t = bson_iter_type(&doc_iter)
+                if ftype == _BsonArrowTypes.int32:
+                    if value_t == BSON_TYPE_INT32:
```
ShaneHarvey, line 39: Should int32 also accept BSON_TYPE_BOOL like int64 and double? There should be tests for each of these conversion rules. 

Will these lossy conversion rules cause problems when we later add support for writing (BSON bool -> int64 -> BSON int64)?

bindings/python/pymongoarrow/bson.pyi
```diff
@@ -0,0 +1,66 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def process_bson_stream(bson_stream, context):
+    cdef const uint8_t* docstream = <const uint8_t *>bson_stream
+    cdef size_t length = <size_t>PyBytes_Size(bson_stream)
+    cdef bson_reader_t* stream_reader = bson_reader_new_from_data(docstream, length)
```
ShaneHarvey, line 19: Does bson_reader_new_from_data need to be destroyed?

bindings/python/pymongoarrow/builders.pyi
```diff
@@ -12,9 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Cython compiler directives
-# distutils: language=c++
-# cython: language_level=3
```
ShaneHarvey, line 17: Was this not needed or causing problems?

bindings/python/pymongoarrow/context.py
```diff
@@ -0,0 +1,59 @@
...
+
+_TYPE_TO_BUILDER_CLS = {
+    _BsonArrowTypes.int32: Int32Builder,
+    _BsonArrowTypes.int64: Int64Builder,
+    _BsonArrowTypes.double: DoubleBuilder,
+    _BsonArrowTypes.datetime: DatetimeBuilder
+}
+
+
+class PyMongoArrowContext:
```
ShaneHarvey, line 27: It would be helpful to add docstrings to describe the inputs/outputs/classes in this PR.

bindings/python/pymongoarrow/bson.pyi
```diff
@@ -0,0 +1,66 @@
...
+
+    builder_map = context.builder_map
+    type_map = context.type_map
+
+    while doc != NULL:
+        if not bson_iter_init(&doc_iter, doc):
+            raise RuntimeError
+        while bson_iter_next(&doc_iter):
+            key = bson_iter_key(&doc_iter)
+            if key in builder_map:
```
ShaneHarvey, line 34: What happens when a key is missing from one of docs? I think we should append a Null but it seems like we would skip it altogether. For example what happens in this scenario:
```
docs = [{'_id': 1, 'data': 10},
        {'_id': 2},
        {'_id': 3, 'data': 30},
        {'data': 40}]
schema = Schema({'_id': int32(), 'data': int64()})
...
```

ShaneHarvey (changes requested):
Neat!

bindings/python/pymongoarrow/schema.py
prashantmital: Done.

bindings/python/pymongoarrow/types.py
prashantmital: Good point. Done.

bindings/python/pymongoarrow/builders.pyi
ShaneHarvey: Gotcha.

bindings/python/pymongoarrow/bson.pyi
ShaneHarvey: bson_reader_destroy needs to be called in a finally block to avoid leaking after an exception is raised.

bindings/python/pymongoarrow/bson.pyi
```diff
@@ -0,0 +1,68 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def process_bson_stream(bson_stream, context):
+    cdef const uint8_t* docstream = <const uint8_t *>bson_stream
+    cdef size_t length = <size_t>PyBytes_Size(bson_stream)
+    cdef bson_reader_t* stream_reader = bson_reader_new_from_data(docstream, length)
+    cdef cbool reached_eof = False
+    cdef const bson_t * doc = bson_reader_read(stream_reader, & reached_eof)
```
ShaneHarvey, line 21: We're missing error handling for the `bson_reader_read` calls. See http://mongoc.org/libbson/current/bson_reader_read.html

bindings/python/pymongoarrow/bson.pyi
ShaneHarvey: In this PR or a follow up PR?

bindings/python/pymongoarrow/bson.pyi
ShaneHarvey: Oh I see, the issue is that libbson provides bson_iter_as_int64 and bson_iter_as_double which does the casting. There is no bson_iter_as_int32 for us to use here so this code SGTM as is.

bindings/python/pymongoarrow/context.py
ShaneHarvey: It would be helpful to do the same for the Schema class too.

bindings/python/pymongoarrow/context.py
```diff
@@ -0,0 +1,75 @@
...
+            type_map[encoded_fname] = ftype
+        return cls(schema, builder_map, type_map)
+
+    def __contains__(self, item):
+        if isinstance(item, str):
+            return item in self.schema
+        elif isinstance(item, bytes):
+            return item in self.builder_map
+        else:
+            raise TypeError('context keys are either str or bytes type')
```
ShaneHarvey, line 67: `__contains__` seems to be unused. Can we remove it?

ShaneHarvey (changes requested):

bindings/python/pymongoarrow/bson.pyi
prashantmital: Done.

bindings/python/pymongoarrow/context.py
prashantmital: Done

bindings/python/pymongoarrow/context.py
prashantmital: Done.

bindings/python/pymongoarrow/bson.pyi
prashantmital: Done.

bindings/python/pymongoarrow/bson.pyi
prashantmital: Done.

bindings/python/pymongoarrow/bson.pyi
```diff
@@ -13,56 +13,69 @@
...
-                        builder.append(bson_iter_date_time(&doc_iter))
-                    else:
-                        builder.append_null()
-                else:
-                    raise TypeError('unknown ftype {}'.format(ftype))
-        doc = bson_reader_read(stream_reader, &reached_eof)
-
-    bson_reader_destroy(stream_reader)
+                        raise TypeError('unknown ftype {}'.format(ftype))
+            count += 1
```
ShaneHarvey, line 75: Nice! Minor nit: Suggest adding a comment here:
`# Append null to account for any missing field(s)`

bindings/python/pymongoarrow/bson.pyi
```diff
@@ -13,56 +13,69 @@
...
-                    else:
-                        builder.append_null()
-                elif ftype == _BsonArrowTypes.int64:
-                    if (value_t == BSON_TYPE_INT64 or
-                            value_t == BSON_TYPE_BOOL or
-                            value_t == BSON_TYPE_DOUBLE or
-                            value_t == BSON_TYPE_INT32):
-                        builder.append(bson_iter_as_int64(&doc_iter))
+    try:
+        while doc != NULL:
```
ShaneHarvey, line 38: Nit: I suggest putting the first call to bson_reader_read_safe inside the try/except. We could even refactor to have a single call instead of 2:
```
    try:
        while True:
            doc = bson_reader_read_safe(stream_reader, &reached_eof)
            if doc == NULL:
                break
            if not bson_iter_init(&doc_iter, doc):....
```

bindings/python/pymongoarrow/bson.pyi
```diff
@@ -13,56 +13,69 @@
...
+            if not bson_iter_init(&doc_iter, doc):
+                raise RuntimeError
```
ShaneHarvey, line 40: Is there a better error we can use here?

bindings/python/pymongoarrow/builders.pyi
```diff
@@ -13,27 +13,26 @@
...
         for value in values:
             self.append(value)
 
-    @property
-    def null_count(self):
-        return self.builder.get().null_count()
-
-    def __len__(self):
-        return self.builder.get().length()
-
```
ShaneHarvey, line 30: Why was this required?

bindings/python/test/test_bson.py
```diff
@@ -22,34 +22,73 @@
...
-                             'data': int64()})
-        cls.bson_stream = payload
+
+class TestInvalidBsonToArrowConversion(TestBsonToArrowConversionBase):
+    @staticmethod
+    def _generate_payload(doclist):
+        payload = b''
+        for doc in doclist:
+            payload += encode(doc)
+        return payload[:-2]
```
ShaneHarvey, line 81: Can this be replaced with `super()._generate_payload()[:-2]`?

bindings/python/test/test_bson.py
```diff
@@ -22,34 +22,73 @@
...
-        as_dict = defaultdict(list)
-        for doc in docs:
-            for key, value in doc.items():
-                as_dict[key].append(value)
+    def test_with_nulls(self):
+        docs = [{'_id': 1, 'data': 10},
+                {'_id': 2, 'data': 20},
+                {'_id': 3},
+                {'_id': 4, 'data': 40},
+                {'foo': 1}]
```
ShaneHarvey, line 67: Suggest adding a totally empty document as well.

bindings/python/pymongoarrow/bson.pyi
```diff
@@ -13,56 +13,69 @@
 # limitations under the License.
 
 
+cdef const bson_t* bson_reader_read_safe(bson_reader_t* stream_reader, cbool* reached_eof) except? NULL:
+    cdef const bson_t* doc = bson_reader_read(stream_reader, reached_eof)
+    if doc == NULL and dereference(reached_eof) == False:
+        raise RuntimeError("Could not read BSON stream")
```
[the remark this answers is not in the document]
prashantmital: I think there would definitely be better errors for us to use here... Which one do you think we should use? I think we can safely defer the decision to a separate ticket - if you are OK with this, let me know and I will create one.

bindings/python/pymongoarrow/bson.pyi
prashantmital: Done.

bindings/python/pymongoarrow/bson.pyi
prashantmital: Nice! Done.

bindings/python/pymongoarrow/bson.pyi
prashantmital: See comment above - I think we can defer to a separate ticket. PyMongoArrow should probably expose some exceptions of its own

bindings/python/pymongoarrow/builders.pyi
prashantmital: There are some weird errors that crop up when you try to use C-namespace things from a parent class method in a child `cdef class`. The easiest workaround was to simply equip each `cdef class` with its own instance of this method.

bindings/python/test/test_bson.py
prashantmital: Doesn't seem to work.. Some strange interaction with `super` and `staticmethod` I think?
```
======================================================================
ERROR: test_simple (test.test_bson.TestInvalidBsonToArrowConversion)
----------------------------------------------------------------------
  File "/Users/pmital/Developer/mongo-arrow-github/bindings/python/test/test_bson.py", line 95, in test_simple
    self._run_test(docs, as_dict)
  File "/Users/pmital/Developer/mongo-arrow-github/bindings/python/test/test_bson.py", line 40, in _run_test
    payload = type(self)._generate_payload(doclist)
  File "/Users/pmital/Developer/mongo-arrow-github/bindings/python/test/test_bson.py", line 78, in _generate_payload
    return super()._generate_payload()[:-2]
TypeError: super(type, obj): obj must be an instance or subtype of type

----------------------------------------------------------------------
```

bindings/python/test/test_bson.py
prashantmital: Done.

bindings/python/test/test_bson.py
ShaneHarvey: Apparently you need to do:
```
super(TestInvalidBsonToArrowConversion, TestInvalidBsonToArrowConversion)._generate_payload()[:-2]?
```
https://stackoverflow.com/a/26807879/8637283

bindings/python/pymongoarrow/bson.pyi
ShaneHarvey: I think we should at least add an error message to the runtimeerror here.

bindings/python/pymongoarrow/bson.pyi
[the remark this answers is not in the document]
ShaneHarvey: Pymongo's InvalidBSON error makes sense to me. Feel free to open a new ticket for it.

bindings/python/pymongoarrow/builders.pyi
ShaneHarvey: Thanks.

bindings/python/pymongoarrow/bson.pyi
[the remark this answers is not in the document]
prashantmital: Opened PYTHON-2595

bindings/python/pymongoarrow/bson.pyi
prashantmital: Done.

bindings/python/test/test_bson.py
prashantmital: Done.

ShaneHarvey (approved):

PR #4, opened 2021-03-09
Public API for Arrow

prashantmital:
# To Do

- [x] Add `find` API
- [x] Test `find` API
- [x] Add `aggregate` API
- [x] Test `aggregate` API
- [ ] Enable automated testing via GitHub Actions

### DIFF
--- a/bindings/python/pymongoarrow/api.py
+++ b/bindings/python/pymongoarrow/api.py
@@ -11,6 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import warnings
+
 from pymongoarrow.context import PyMongoArrowContext
 from pymongoarrow.lib import process_bson_stream
 from pymongoarrow.schema import Schema
@@ -39,8 +41,16 @@ def find_arrow_all(collection, query, schema, **kwargs):
       An instance of class:`pyarrow.Table`.
     """
     context = PyMongoArrowContext.from_schema(schema)
+
+    for opt in ('session', 'cursor_type', 'session'):
+        if kwargs.pop(opt, None):
+            warnings.warn(
+                'Ignoring option {!r} as it is not supported by '
+                'PyMongoArrow'.format(opt), UserWarning, stacklevel=2)
+
     raw_batch_cursor = collection.find_raw_batches(
         query, **kwargs)
     for batch in raw_batch_cursor:
         process_bson_stream(batch, context)
+
     return context.finish()
--- a/bindings/python/pymongoarrow/api.py
+++ b/bindings/python/pymongoarrow/api.py
@@ -35,11 +35,11 @@ def find_arrow_all(collection, query, *, schema, **kwargs):
     """
     context = PyMongoArrowContext.from_schema(schema)
 
-    for opt in ('session', 'cursor_type', 'session', 'projection'):
+    for opt in ('session', 'cursor_type', 'projection'):
         if kwargs.pop(opt, None):
             warnings.warn(
-                'Ignoring option {!r} as it is not supported by '
-                'PyMongoArrow'.format(opt), UserWarning, stacklevel=2)
+                f'Ignoring option {opt!r} as it is not supported by '
+                'PyMongoArrow', UserWarning, stacklevel=2)
 
     kwargs['projection'] = schema._get_projection()
     raw_batch_cursor = collection.find_raw_batches(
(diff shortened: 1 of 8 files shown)


bindings/python/pymongoarrow/api.py
```diff
@@ -0,0 +1,87 @@
...
+
+    Additional keyword-arguments passed to this method will be passed
+    directly to the underlying ``find`` operation.
+
+    :Returns:
+      An instance of class:`pyarrow.Table`.
+    """
+    context = PyMongoArrowContext.from_schema(schema)
+
+    for opt in ('session', 'cursor_type', 'session', 'projection'):
```
ShaneHarvey, line 38: Session appears twice here.

bindings/python/pymongoarrow/api.py
```diff
@@ -0,0 +1,87 @@
...
+        if kwargs.pop(opt, None):
+            warnings.warn(
+                'Ignoring option {!r} as it is not supported by '
+                'PyMongoArrow'.format(opt), UserWarning, stacklevel=2)
```
ShaneHarvey, line 42: Is there any particular reason we're using `"".format()` instead of format strings like `f'..{opt}...'`?

bindings/python/pymongoarrow/api.py
```diff
@@ -0,0 +1,87 @@
...
+    Additional keyword-arguments passed to this method will be passed
+    directly to the underlying ``aggregate`` operation.
+
+    :Returns:
+      An instance of class:`pyarrow.Table`.
+    """
+    context = PyMongoArrowContext.from_schema(schema)
+
+    if pipeline and ("$out" in pipeline[-1] or "$merge" in pipeline[-1]):
+        raise RuntimeError(
```
ShaneHarvey, line 72: Should be `ValueError` imo since the `pipeline` value is not valid.

bindings/python/test/test_arrow.py
```diff
@@ -0,0 +1,117 @@
...
+import unittest
+import unittest.mock as mock
+
+from pyarrow import int32, int64, schema as ArrowSchema, Table
+from pymongo import WriteConcern, DESCENDING
+import pymongo
+from pymongoarrow import aggregate_arrow_all, find_arrow_all, Schema
+
+from test import client_context
+from test.utils import WhiteListEventListener
```
ShaneHarvey, line 24: WhiteListEventListener -> AllowListEventListener

bindings/python/test/__init__.py
```diff
@@ -11,3 +11,40 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import pymongo
+
+
+class ClientContext:
+    def __init__(self, host='localhost', port=27017, client_kwargs=None):
```
ShaneHarvey, line 18: `client_kwargs` is never used. Can we remove it?

bindings/python/test/test_arrow.py
```diff
@@ -0,0 +1,117 @@
...
+    def setUpClass(cls):
+        if not client_context.connected:
+            raise unittest.SkipTest("cannot connect to MongoDB")
+        cls.cmd_listener = WhiteListEventListener('find', 'aggregate')
+        cls.getmore_listener = WhiteListEventListener('getMore')
+        cls.client = client_context.get_client(
+            event_listeners=[cls.getmore_listener, cls.cmd_listener])
+        cls.schema = Schema({'_id': int32(), 'data': int64()})
+        cls.coll = cls.client.pymongoarrow_test.get_collection(
+            'test', write_concern=WriteConcern(w='majority'))
```
ShaneHarvey, line 38: Why use `w=majority` explicitly?

bindings/python/pymongoarrow/schema.py
```diff
@@ -68,6 +68,12 @@ def _normalize_sequence(sequence):
                 normed[fname] = _normalize_typeid(ftype, fname)
         return normed
 
+    def _get_projection(self):
+        projection = {'_id': False}
```
ShaneHarvey, line 72: We never seem to test that `_id` is omitted by default. Should we add a test for _get_projection in test_schema.py?

bindings/python/pymongoarrow/api.py
prashantmital: Done.

bindings/python/pymongoarrow/api.py
prashantmital: None other than I wasn't fully familiar with f-strings. Changed :)

bindings/python/pymongoarrow/api.py
prashantmital: Changed.

bindings/python/test/test_arrow.py
prashantmital: Oof! Thanks for catching that.

bindings/python/test/__init__.py
prashantmital: Removed.

bindings/python/test/test_arrow.py
prashantmital: Force of habit. Any reason we shouldn't be? 
I thought it was good practice to use it to prevent some weird race condition where the first test doesn't see all the documents written during setup because they haven't propagated to a majority of nodes...

bindings/python/pymongoarrow/schema.py
prashantmital: Instead of unit testing that private method, I have added integration tests that check the `_id` field is omitted from server responses.

bindings/python/test/test_arrow.py
ShaneHarvey: That case only happens when you try to read your own writes from secondary and we added causal consistency to address that problem (w:majority doesn't fix this). The real reason to use w=majority is to avoid rollbacks; that is where the primary writes some data locally, responds ok:1, the client thinks the write is persisted, then before the write can propagate to a majority of nodes a new primary is elected and the write is later rolled back. I've never seen rollback happen in a driver test run ever so I don't think it's much of a concern but it is best practice. Feel free to keep it.

ShaneHarvey (commented):
Looks good so far. Are you planning to add the Actions test suite?

bindings/python/test/test_arrow.py
prashantmital: 👍🏽

prashantmital:
I was hoping to add the GH Actions testing later as this is right now ready to merge.

ShaneHarvey (approved):

PR #5, opened 2021-03-16
Add hook for monkey-patching pymongo's Collection

prashantmital:
Depends on #4

### DIFF
--- a/bindings/python/pymongoarrow/__init__.py
+++ b/bindings/python/pymongoarrow/__init__.py
@@ -29,14 +29,3 @@ if _parse_version(libbson_version) < _parse_version(_MIN_LIBBSON_VERSION):
     raise ImportError(
         "Expected libbson version {} or greater, found {}}".format(
             _MIN_LIBBSON_VERSION, libbson_version))
-
-
-from pymongoarrow.api import aggregate_arrow_all, find_arrow_all
-from pymongoarrow.schema import Schema
-
-
-__all__ = [
-    'aggregate_arrow_all',
-    'find_arrow_all',
-    'Schema'
-]
--- a/bindings/python/pymongoarrow/api.py
+++ b/bindings/python/pymongoarrow/api.py
@@ -26,18 +26,12 @@ __all__ = [
 ]
 
 
-def patch_function(**kwds):
-    def decorator(func):
-        @functools.wraps(func)
-        def wrapped(*args, **kwargs):
-            return func(*args, **kwargs)
-        for attr, val in kwds.items():
-            setattr(wrapped, attr, val)
-        return wrapped
-    return decorator
+_PATCH_METHODS = [
+    'aggregate_arrow_all',
+    'find_arrow_all',
+]
 
 
-@patch_function(__target__='Collection')
 def find_arrow_all(collection, query, *, schema, **kwargs):
     """Method that returns the results of a find query as a
     :class:`pyarrow.Table` instance.
@@ -71,7 +65,6 @@ def find_arrow_all(collection, query, *, schema, **kwargs):
     return context.finish()
 
 
-@patch_function(__target__='Collection')
 def aggregate_arrow_all(collection, pipeline, *, schema, **kwargs):
     """Method that returns the results of an aggregation pipeline as a
     :class:`pyarrow.Table` instance.
--- a/bindings/python/pymongoarrow/api.py
+++ b/bindings/python/pymongoarrow/api.py
@@ -11,12 +11,33 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import functools
 import warnings
 
 from pymongoarrow.context import PyMongoArrowContext
 from pymongoarrow.lib import process_bson_stream
+from pymongoarrow.schema import Schema
 
 
+__all__ = [
+    'aggregate_arrow_all',
+    'find_arrow_all',
+    'Schema'
+]
+
+
+def patch_function(**kwds):
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapped(*args, **kwargs):
+            return func(*args, **kwargs)
+        for attr, val in kwds.items():
+            setattr(wrapped, attr, val)
+        return wrapped
+    return decorator
+
+
+@patch_function(__target__='pymongo.collection.Collection')
 def find_arrow_all(collection, query, *, schema, **kwargs):
     """Method that returns the results of a find query as a
     :class:`pyarrow.Table` instance.
@@ -50,6 +71,7 @@ def find_arrow_all(collection, query, *, schema, **kwargs):
     return context.finish()
 
 
+@patch_function(__target__='pymongo.collection.Collection')
 def aggregate_arrow_all(collection, pipeline, *, schema, **kwargs):
     """Method that returns the results of an aggregation pipeline as a
     :class:`pyarrow.Table` instance.
--- a/bindings/python/pymongoarrow/api.py
+++ b/bindings/python/pymongoarrow/api.py
@@ -37,7 +37,7 @@ def patch_function(**kwds):
     return decorator
 
 
-@patch_function(__target__='pymongo.collection.Collection')
+@patch_function(__target__='Collection')
 def find_arrow_all(collection, query, *, schema, **kwargs):
     """Method that returns the results of a find query as a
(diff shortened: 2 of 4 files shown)


bindings/python/pymongoarrow/monkey.py
```diff
@@ -0,0 +1,33 @@
...
+]
+
+
+def patch_all():
+    for method in (aggregate_arrow_all, find_arrow_all):
+        if not hasattr(method, "__target__"):
+            raise ValueError(
+                "cannot patch method with unset '__target__' attribute")
+
+        target_spec = method.__target__
```
ShaneHarvey, line 29: This seems a bit overly complex. Is there any downside to this approach:
```python
from pymongo.collection import Collection
for method in (aggregate_arrow_all, find_arrow_all):
    setattr(Collection, method.__name__, method)

File: bindings/python/pymongoarrow/__init__.py [hunk 2]
```diff
@@ -29,14 +29,3 @@ def _parse_version(version):
...
-
-from pymongoarrow.api import aggregate_arrow_all, find_arrow_all
-from pymongoarrow.schema import Schema
-
-
-__all__ = [
-    'aggregate_arrow_all',
-    'find_arrow_all',
-    'Schema'
-]
```
ShaneHarvey, line 42: Why remove this?

bindings/python/test/test_arrow.py
```diff
@@ -147,3 +153,24 @@ def test_aggregate_omits_id_if_not_in_schema(self):
             if op_name == '$project':
                 self.assertFalse(stage[op_name]['_id'])
                 break
+
+
+class TestArrowExplicitApi(TestArrowApiMixin, unittest.TestCase):
```
ShaneHarvey, line 158: Isn't it overkill to run all the test cases on both the explicit and patched api given that they are the exact same method? I think it would make more sense just to run all the tests with the explicit api and then have one test for the patched methods (or vice versa).

bindings/python/pymongoarrow/__init__.py
prashantmital: Defining `__all__` here was defining the public API for the entire module. This was causing PyCharm to basically throw warnings for every single import in the project because it construed it as me trying to use non-public members of all submodules.

bindings/python/pymongoarrow/monkey.py
prashantmital: The downside is that we cannot pass the module name/path to be imported as a variable to the vanilla `import` statement. This will become a factor if/when we decide to patch other classes (e.g. `Database.aggregate_arrow_all`). I think this is a more general solution.

bindings/python/test/test_arrow.py
prashantmital: Yes, it kind of is overkill but honestly, the cost is so low right now... It takes <1 second to run this entire test file on my laptop...

bindings/python/pymongoarrow/monkey.py
prashantmital: If we go with your solution we might as well also get rid of the whole business with the decorator and hardcode everything that needs to be patched. I am fine with that solution, but I thought this was more elegant as it allows us to simply apply the decorator to any function that needs to be patched onto one of PyMongo's APIs. I think it will come in really handy as we add more functionality to this project (e.g. `Collection.find_arrow_batches`)

prashantmital:
I made one more change which will result in all methods with the `__target__` attribute being automatically patched onto the target classes.

bindings/python/pymongoarrow/__init__.py
prashantmital: I think eventually we will need to define some sensible `__all__` here but it can wait while the project is still under development.

bindings/python/pymongoarrow/monkey.py
ShaneHarvey: I would strongly prefer the _less_ general approach since that's all we need right now. Also if we wanted to patch Database that would also be quite straightforward:
```
from pymongo.database import Database
Database.aggregate_arrow_all = aggregate_arrow_all
```

bindings/python/pymongoarrow/__init__.py
ShaneHarvey: Interesting, thanks for explaining.

bindings/python/test/test_arrow.py
ShaneHarvey: Okay, let's keep this in mind for future testing.

bindings/python/pymongoarrow/monkey.py
prashantmital: Changed to use what you suggested.

ShaneHarvey: Thanks, keeping this simple makes the code easier to read and lessens the risk of bugs.

bindings/python/pymongoarrow/api.py
```diff
@@ -11,12 +11,33 @@
...
+        @functools.wraps(func)
+        def wrapped(*args, **kwargs):
+            return func(*args, **kwargs)
+        for attr, val in kwds.items():
+            setattr(wrapped, attr, val)
+        return wrapped
+    return decorator
+
+
+@patch_function(__target__='Collection')
```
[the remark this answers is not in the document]
prashantmital: I think we need to keep it since this helps us identify which functions in `pymongoarrow.api` need to be patched at all.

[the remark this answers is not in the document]
prashantmital: I can replace it with some kind of sentinel tag if you'd prefer, but I'd really rather not have to make changes in `monkey.py` each time a new function that needs to be patched somewhere is added.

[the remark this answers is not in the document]
ShaneHarvey: Isn't it much simpler for the patch method to do the patching?:
```python
def patch():
    from pymongo.collection import Collection
    for method in (aggregate_arrow_all, find_arrow_all):
        setattr(Collection, method.__name__, method)
```

Why would we need a generalized patching solution for such a small problem that's easily solved in 3 lines?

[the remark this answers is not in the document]
prashantmital: > Why would we need a generalized patching solution for such a small problem that's easily solved in 3 lines?

The 'solution' makes it easier to add more methods that are to be patched onto Collection. In the above example, I first need to define the desired method in `pymongoarrow.api`, then import it in `pymongoarrow.monkey` and finally add it to the `for`-loop. Under the current approach, new methods that require patching onto `Collection` just need to be declared with a `@patch` decorator.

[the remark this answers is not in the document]
prashantmital: I'll change it to what you suggested if you'd like but I am really not a fan of needing to touch 2 files every time we add a method that also needs to be patched.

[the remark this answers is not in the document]
ShaneHarvey: Again I much prefer simplicity here (avoiding __import__, avoiding decorators). It doesn't seem like much of a problem to touch 2 files given that we're only planning to add 6 methods and that is a one time cost. The alternative is a continuous cost of reading the more complex patch decorator code (for everyone who ever reads the project).

I think the best of both worlds would be in api.py:
```python
# The methods to patch onto pymongo's Collection class
_PATCH_FUNCS = (aggregate_arrow_all, find_arrow_all)
```
Then in monkey.py:
```python
def patch_all():
    from pymongo.collection import Collection
    for method in _PATCH_FUNCS:
        setattr(Collection, method.__name__, method)
```

Easy to read, easy to understand, concise, no duplication.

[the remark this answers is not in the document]
prashantmital: Done.

bindings/python/pymongoarrow/api.py
```diff
@@ -11,10 +11,25 @@
...
+
+
+__all__ = [
+    'aggregate_arrow_all',
+    'find_arrow_all',
+    'Schema'
+]
+
+
+_PATCH_METHODS = [
```
prashantmital, line 29: I listed these as strings instead of function handles because the names are not yet declared at this point in the file. I didn't particularly like the idea of declaring `_PATCH_METHODS` at the bottom of the file so I did this.

ShaneHarvey (approved):

bindings/python/pymongoarrow/api.py
ShaneHarvey: Cool, I think either way is fine.

PR #6, opened 2021-03-18
Public API for Pandas

prashantmital:
Though the design calls for it, I think we should drop the `na_value` field because pandas has great utilities for handling missing values - https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.fillna.html
If we did add `na_value` support it would basically be a pass-through which doesn't make too much sense to me... It might be better handled by really good docs.

### DIFF
--- a/bindings/python/pymongoarrow/api.py
+++ b/bindings/python/pymongoarrow/api.py
@@ -100,3 +100,58 @@ def aggregate_arrow_all(collection, pipeline, *, schema, **kwargs):
         process_bson_stream(batch, context)
 
     return context.finish()
+
+
+def find_pandas_all(collection, query, *, schema, na_value=None, **kwargs):
+    """Method that returns the results of a find query as a
+    :class:`pandas.DataFrame` instance.
+
+    :Parameters:
+      - `collection`: Instance of :class:`~pymongo.collection.Collection`.
+        against which to run the ``find`` operation.
+      - `query`: A mapping containing the query to use for the find operation.
+      - `schema`: Instance of :class:`~pymongoarrow.schema.Schema`.
+      - `na_value`: Scalar or mapping containing the value to use when a
+        field doesn't exist in a query result or when it exists but is of
+        a type that cannot be safely cast to the target type. If this is a
+        scalar, the missing value applies to all fields. If this is a mapping,
+        the keys are field names and the values are the scalars to use for
+        missing entries for the named field.
+
+    Additional keyword-arguments passed to this method will be passed
+    directly to the underlying ``find`` operation.
+
+    :Returns:
+      An instance of class:`pandas.DataFrame`.
+    """
+    arrow_table = find_arrow_all(collection, query, schema=schema, **kwargs)
+    pandas_table = arrow_table.to_pandas(split_blocks=True, self_destruct=True)
+    return pandas_table
+
+
+def aggregate_pandas_all(collection, pipeline, *, schema, na_value=None, **kwargs):
+    """Method that returns the results of an aggregation pipeline as a
+    :class:`pandas.DataFrame` instance.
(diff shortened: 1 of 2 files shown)


bindings/python/pymongoarrow/api.py
```diff
@@ -85,3 +86,58 @@ def aggregate_arrow_all(collection, pipeline, *, schema, **kwargs):
...
+def find_pandas_all(collection, query, *, schema, na_value=None, **kwargs):
+    """Method that returns the results of a find query as a
+    :class:`pandas.DataFrame` instance.
+
+    :Parameters:
+      - `collection`: Instance of :class:`~pymongo.collection.Collection`.
+        against which to run the ``find`` operation.
+      - `query`: A mapping containing the query to use for the find operation.
+      - `schema`: Instance of :class:`~pymongoarrow.schema.Schema`.
+      - `na_value`: Scalar or mapping containing the value to use when a
```
ShaneHarvey, line 100: Let's remove na_value.

bindings/python/pymongoarrow/api.py
```diff
@@ -85,3 +86,58 @@ def aggregate_arrow_all(collection, pipeline, *, schema, **kwargs):
...
+def aggregate_pandas_all(collection, pipeline, *, schema, na_value=None, **kwargs):
+    """Method that returns the results of an aggregation pipeline as a
+    :class:`pandas.DataFrame` instance.
+
+    :Parameters:
+      - `collection`: Instance of :class:`~pymongo.collection.Collection`.
+        against which to run the ``find`` operation.
+      - `pipeline`: A list of aggregation pipeline stages.
+      - `schema`: Instance of :class:`~pymongoarrow.schema.Schema`.
+      - `na_value`: Scalar or mapping containing the value to use when a
```
ShaneHarvey, line 127: Let's remove na_value.

bindings/python/pymongoarrow/api.py
```diff
@@ -85,3 +86,58 @@ def aggregate_arrow_all(collection, pipeline, *, schema, **kwargs):
...
+
+    Additional keyword-arguments passed to this method will be passed
+    directly to the underlying ``aggregate`` operation.
+
+    :Returns:
+      An instance of class:`pandas.DataFrame`.
+    """
+    arrow_table = aggregate_arrow_all(
+        collection, pipeline, schema=schema, **kwargs)
+    pandas_table = arrow_table.to_pandas(split_blocks=True, self_destruct=True)
```
ShaneHarvey, line 142: Can you add a comment explaining `split_blocks=True, self_destruct=True`? Perhaps it would be helpful to add a `_to_pandas` helper method to avoid duplicating the comment.

ShaneHarvey (commented):
I agree that we should let users call fillna() directly instead of adding syntactic sugar for it. Let's record the rationale for why we made this decision somewhere in Jira.

bindings/python/pymongoarrow/api.py
```diff
@@ -100,3 +101,54 @@ def aggregate_arrow_all(collection, pipeline, *, schema, **kwargs):
         process_bson_stream(batch, context)
 
     return context.finish()
+
+
+def _arrow_to_pandas_efficient(arrow_table):
```
ShaneHarvey, line 106: nit: let's drop the "_efficient" since hopefully all our code is as efficient as it needs to be and I don't see us adding a _arrow_to_pandas_inefficient :)

ShaneHarvey (approved):
LGTM with a minor nit

bindings/python/pymongoarrow/api.py
prashantmital: lol

PR #12, opened 2021-04-09
Arrow CI Improvements

prashantmital:


bindings/python/release.sh
```diff
@@ -7,26 +7,21 @@ set -o errexit
...
-
 # Compute shared library filename
 if [ "Darwin" = "$(uname -s)" ]
 then
   SO_EXT='dylib'
 elif [ "Darwin" = "$(uname -s)" ]
 then
   SO_EXT='so'
-  export MACOSX_DEPLOYMENT_TARGET="10.15"
+  export MACOSX_DEPLOYMENT_TARGET="10.9"
```
prashantmital, line 17: This gets rid of the `object file . . . was built for newer OSX version (10.xx) than being linked (10.yy)` warnings. I haven't been able to produce an error, but I think our extensions will fail without this because we use C++ (it doesn't affect C so extensions that are C-only are not affected).

ShaneHarvey (approved):

PR #13, opened 2021-04-09
PYTHON-2638 Datetime support

prashantmital:


bindings/python/pymongoarrow/builders.pyi
```diff
@@ -123,10 +123,10 @@ cdef class DatetimeBuilder(_ArrayBuilderBase):
     def __cinit__(self, TimestampType dtype=timestamp('ms'),
                   MemoryPool memory_pool=None):
         cdef CMemoryPool* pool = maybe_unbox_memory_pool(memory_pool)
-        if dtype in (timestamp('us'), timestamp('ns')):
-            raise ValueError("Microsecond resolution temporal type is not "
-                             "suitable for use with MongoDB's UTC datetime "
-                             "type which has resolution of milliseconds.")
+        if dtype.unit in ('s', 'us', 'ns'):
```
ShaneHarvey, line 126: Should this be `if dtype.unit != 'ms':`?

prashantmital:
Some funkiness with what arrow does with timestamps is resulting in a fixed offset in the round tripped date times... I will investigate and fix the builder tests.

bindings/python/test/test_builders.py
```diff
@@ -54,33 +54,40 @@ def test_default_unit(self):
         builder = DatetimeBuilder()
         self.assertEqual(builder.unit, timestamp('ms'))
 
-    def _test_simple(self, tstamp_units, kwarg_name):
-        builder = DatetimeBuilder(dtype=timestamp(tstamp_units))
-        datetimes = [datetime(1970, 1, 1) + timedelta(**{kwarg_name: k*100})
+    def _to_int(self, dt):
+        # datetime.timestamp assumes local time so we must explicitly
+        # specify tzinfo=timezone.utc to obtain POSIX time
+        return int(dt.replace(tzinfo=timezone.utc).timestamp() * 1000)
```
ShaneHarvey, line 60: Is this timezone stuff still relevant given we use `datetime.utcnow` in the tests?

ShaneHarvey (commented):
What about the TODO in datetime_to_int64?
```
    # TODO: make this function aware of datatype.timezone()
```

This seems like a problem.

bindings/python/test/test_builders.py
prashantmital: Yeah, unfortunately. the `.timestamp()` method always assumes times to be local. See the docs - https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp

ShaneHarvey: FYI this is what we do in pymongo:
```python
def _datetime_to_millis(dtm):
    """Convert datetime to milliseconds since epoch UTC."""
    if dtm.utcoffset() is not None:
        dtm = dtm - dtm.utcoffset()
    return int(calendar.timegm(dtm.timetuple()) * 1000 +
               dtm.microsecond // 1000)
```
https://github.com/mongodb/mongo-python-driver/blob/3.11.3/bson/__init__.py#L902-L907

prashantmital:
Actually that is dead code... I originally wrote that function to support other granularities for timestamps but eventually decided to not support that. In our case since we only support reading from Mongo/BSON to Arrow, there is no need for this method.

bindings/python/pymongoarrow/bson.pyi
```diff
@@ -29,7 +29,7 @@ def process_bson_stream(bson_stream, context):
     cdef bson_iter_t doc_iter
     cdef const char* key
     cdef bson_type_t value_t
-    cdef uint64_t count = 0
+    cdef Py_ssize_t count = 0
```
prashantmital, line 32: This is an unrelated change that eliminates this warning:
```
pymongoarrow/lib.cpp:5187:34: warning: comparison of integers of different signs: 'Py_ssize_t' (aka 'long') and 'uint64_t' (aka 'unsigned long long') [-Wsign-compare]
        __pyx_t_12 = ((__pyx_t_9 != __pyx_v_count) != 0);
```

bindings/python/test/test_builders.py
ShaneHarvey: I think there is a misunderstanding here. The Python docs say:

> **Naive datetime** instances are assumed to represent local time...
> For **aware datetime** instances, the return value is computed as:
>    (dt - datetime(1970, 1, 1, tzinfo=timezone.utc)).total_seconds()

As it stands the current `_to_int` doesn't work correctly for aware datetimes. `replace(tzinfo=timezone.utc)` truncates the original timezone and replaces it with UTC but doesn't apply the offset. We should use pymongo's method which we know is correct.

prashantmital: Vendored that method from `bson` into the test class.

ShaneHarvey (approved):

PR #14, opened 2021-04-14
PYTHON-2613 Build self-contained manylinux{2010, 2014} wheels

prashantmital:


prashantmital:
While this is working now, it raises a few open questions and things to investigate:

- Why could `g++` not handle the extra whitespace returned by `pkg-config --libs-only-l /path/to/libbson-1.0.pc`? Does this point to an underlying problem?
- Should we really be exporting the output of `pkg-config --libs /path/to/libsson-1.0.pc` to the `LDFLAGS` environment variable? As per GNU Make [docs](http://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html) `LDFLAGS` should only contain what we intend to pass using `-L` to the linker. Library names themselves should be exported as `LDLIBS`.
- Should we also be exporting the `CXXFLAGS` environment variable? As per GNU Make [docs](http://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html) this is the variable used to pass extra args to the C++ compiler not `CFLAGS`.

prashantmital:
> Why could g++ not handle the extra whitespace returned by pkg-config --libs-only-l /path/to/libbson-1.0.pc? Does this point to an underlying problem?

The `pkg-config` output contained plain whitespace as confirmed by the following test:

```
>>>  lnames = query_pkgconfig("pkg-config --libs-only-l libbson/lib/libbson-1.0.pc")
>>> for k in enumerate(lnames):
...     print(k)
... 
(0, '-')
(1, 'l')
(2, 'b')
(3, 's')
(4, 'o')
(5, 'n')
(6, '-')
(7, '1')
(8, '.')
(9, '0')
(10, ' ')
(11, ' ')
>>>
```

This seems to be a bug in `x86_64-linux-gnu-g++` but I am not. certain.

ShaneHarvey (approved):
I don't think there's any bug in the linker. To me it looks like we were accidentally passing `'bson-1.0 '` to Cython which you fixed by passing `'bson-1.0'`.

> Should we really be exporting the output of pkg-config --libs /path/to/libsson-1.0.pc to the LDFLAGS environment variable? As per GNU Make docs LDFLAGS should only contain what we intend to pass using -L to the linker. Library names themselves should be exported as LDLIBS.

Interesting, I did not know about LDLIBS. Let's file this as an improvement ticket in the future.

PR #15, opened 2021-04-20
PYTHON-2649 Add timezone support

prashantmital:
Depends on #13

### DIFF
--- a/bindings/python/pymongoarrow/context.py
+++ b/bindings/python/pymongoarrow/context.py
@@ -11,7 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from pyarrow import Table
+from bson.codec_options import DEFAULT_CODEC_OPTIONS
+
+from pyarrow import timestamp, Table
 from pymongoarrow.lib import Int32Builder, Int64Builder, DoubleBuilder, DatetimeBuilder
 from pymongoarrow.types import _get_internal_typemap, _BsonArrowTypes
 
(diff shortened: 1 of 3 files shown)


prashantmital:
@ShaneHarvey this is ready for your 👀

bindings/python/pymongoarrow/context.py
```diff
@@ -41,21 +43,27 @@ def __init__(self, schema, builder_map, type_map):
...
         builder_map = {}
         type_map = {}
         str_type_map = _get_internal_typemap(schema.typemap)
         for fname, ftype in str_type_map.items():
             builder_cls = _TYPE_TO_BUILDER_CLS[ftype]
             encoded_fname = fname.encode('utf-8')
+            # special-case initializing builders for parameterized types
             if builder_cls == DatetimeBuilder:
                 arrow_type = schema.typemap[fname]
+                if codec_options.tzinfo is not None and arrow_type.tz is None:
```
ShaneHarvey, line 64: Should we need to check codec_options.tz_aware here too? That would make it match pymongo's decoding behavior more closely.

prashantmital: We can, but its not possible to create CodecOptions instances with non-`None` `tzinfo` if `tz_aware` is not also set to `True` so I am not sure it adds much value:
```
CodecOptions(tzinfo=datetime.timezone.utc)
  File "/usr/local/Caskroom/miniconda/base/envs/arrow/lib/python3.7/site-packages/IPython/core/interactiveshell.py", line 3437, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-60-6927577ac152>", line 1, in <module>
    CodecOptions(tzinfo=datetime.timezone.utc)
  File "/usr/local/Caskroom/miniconda/base/envs/arrow/lib/python3.7/site-packages/bson/codec_options.py", line 271, in __new__
    "cannot specify tzinfo without also setting tz_aware=True")
ValueError: cannot specify tzinfo without also setting tz_aware=True
```

ShaneHarvey: Ah thanks for checking that.

ShaneHarvey (approved):

PR #16, opened 2021-04-20
Use requirement files to single-source dependency versions

prashantmital:


.github/workflows/test-python.yml
```diff
@@ -31,12 +31,12 @@ jobs:
...
         run: |
           python -m pip install -U pip wheel
-          pip install 'cython>=0.29' 'pyarrow>=3' 'numpy>=1.16.6' 'pymongo>=3.11,<4' 'pandas>=1.0,<1.2' 'numpy >= 1.16.6'
+          python -m pip install -r requirements/build.txt -r requirements/test.txt
           python -c 'import pyarrow as pa; pa.create_library_symlinks()'
         working-directory: ./bindings/python
       - name: Install pymongoarrow
         run: |
-          LIBBSON_INSTALL_DIR=$(pwd)/libbson python setup.py build_ext --inplace --force
+          LIBBSON_INSTALL_DIR=$(pwd)/libbson python -m pip install -e .
```
ShaneHarvey, line 39: I personally prefer running `build_ext` explicitly here because we get to see the compile output. Right now all we see is:
```
Run LIBBSON_INSTALL_DIR=$(pwd)/libbson python -m pip install -e .
Obtaining file:///home/runner/work/mongo-arrow/mongo-arrow/bindings/python
Requirement already satisfied: pyarrow<3.1,>=3 in /opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages (from pymongoarrow==0.1.0.dev0) (3.0.0)
Collecting pymongo<4,>=3.11
  Downloading pymongo-3.11.3-cp36-cp36m-manylinux2014_x86_64.whl (509 kB)
Requirement already satisfied: numpy<2,>=1.16.6 in /opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages (from pymongoarrow==0.1.0.dev0) (1.19.5)
Installing collected packages: pymongo, pymongoarrow
  Running setup.py develop for pymongoarrow
Successfully installed pymongo-3.11.3 pymongoarrow
```

Does that matter to you?

prashantmital: Good point. Please see latest commit.

prashantmital:
The latest commit:

- includes build requirements in the test requirements file (thanks for the tip)
- install pymongo and pymongoarrow via pip while also displaying compiler output
- displays more verbose output from the suite that include full test names so we aren't left guessing what tests ran and which were skipped

.github/workflows/test-python.yml
ShaneHarvey: The verbose output here is super verbose at 3000+ lines. What if we just recompile explicitly after:
```
LIBBSON_INSTALL_DIR=$(pwd)/libbson python -m pip install -e .
LIBBSON_INSTALL_DIR=$(pwd)/libbson python setup.py build_ext --inplace --force
```

ShaneHarvey (approved):

PR #22, opened 2021-04-26
Calling monkey.patch_all() should patch all API methods

prashantmital:


bindings/python/pymongoarrow/api.py
```diff
@@ -32,6 +32,10 @@
 _PATCH_METHODS = [
     'aggregate_arrow_all',
     'find_arrow_all',
+    'aggregate_pandas_all',
+    'find_pandas_all',
+    'aggregate_numpy_all',
+    'find_numpy_all',
```
ShaneHarvey, line 38: Can we add tests for this?

bindings/python/pymongoarrow/monkey.py
```diff
@@ -11,12 +11,30 @@
...
 __all__ = [
     'patch_all'
 ]
 
 
 def patch_all():
+    """Patch all PyMongoArrow methods into PyMongo.
+
+    This function should be called **before** ``pymongo`` is imported for
+    the first time. Failing to do so might lead to unexpected results.
```
ShaneHarvey, line 25: Is this true? What unexpected results? AFAIK we can patch methods onto classes at anytime.

prashantmital: Removed.

bindings/python/pymongoarrow/api.py
prashantmital: Done.

ShaneHarvey (approved):

PR #24, opened 2021-04-26
Document Type Support

prashantmital:


bindings/python/docs/source/supported_types.rst
```diff
@@ -0,0 +1,34 @@
...
+   * - 64-bit binary floating point
+     - :class:`py.float`, an instance of :meth:`pyarrow.float64`
+   * - 32-bit integer
+     - an instance of :meth:`pyarrow.int32`
+   * - 64-bit integer
+     - :class:`~py.int`, an instance of :meth:`pyarrow.int64`
+   * - UTC datetime
+     - an instance of :class:`~pyarrow.timestamp` with ``ms`` resolution, :class:`py.datetime.datetime`
+   * - 64-bit integer
+     - :class:`~py.int`, :meth:`pyarrow.int64`
```
ShaneHarvey, line 27: `64-bit integer` appears twice.

bindings/python/pymongoarrow/schema.py
```diff
@@ -17,26 +17,29 @@
...
+    To create a schema, provide its constructor a mapping or an iterable
+    containing field names and their expected types, e.g.::
 
-        If ``schema`` is a mapping, each key must be a field name and the
-        corresponding value must be the expected data type of the named field.
-        If ``schema`` is an iterable, it must be comprised entirely of 2-member
-        sub-iterables. The first member of each sub-iterable must be a field
-        name and the second value must be the corresponding data type.
+       schema1 = Schema({'field_1': int, 'field_2': float})
+       schema2 = Schema([('field_1', int), ('field_2', float)])
```
ShaneHarvey, line 26: Why do we support an iterable here instead of only a mapping?

bindings/python/docs/source/supported_types.rst
```diff
@@ -0,0 +1,34 @@
...
+
+.. note:: For more information about BSON types, see the
+   `BSON specification <http://bsonspec.org/spec.html>`_.
+
+.. list-table::
+   :widths: auto
+   :header-rows: 1
+
+   * - BSON Type
+     - Type Identifiers
```
ShaneHarvey, line 17: In the Schema docstring it mentions that bson type are also supported. Should we document them in this table?

bindings/python/pymongoarrow/schema.py
prashantmital: 🤷🏽 . Should we not?

bindings/python/docs/source/supported_types.rst
prashantmital: Ah, right forgot about those.

bindings/python/pymongoarrow/schema.py
ShaneHarvey: It makes it harder to reason about types. Eg `def __init__(self, schema: Mapping[str, Any]):` is simple, `def __init__(self, schema: Union[Mapping[str, Any],Iterable[Tuple[str, Any]]]):` is much more complicated to read.

ShaneHarvey: I would voted to remove it since it doesn't seem to provide any value.

ShaneHarvey (approved):

bindings/python/pymongoarrow/schema.py
ShaneHarvey: Note if you choose to do this we should do it in a separate ticket. This PR LGTM.

prashantmital: Ok sounds good... I agree, we can remove it.

PR #30, opened 2021-04-27
PYTHON-2663 Add manylinux* platform tags to linux wheels

prashantmital:


bindings/python/addtags.py
```diff
@@ -0,0 +1,49 @@
...
+    if not isfile(wheel_path):
+        raise FileNotFoundError('cannot access wheel file %s' % (wheel_path,))
+
+    if not exists(wheel_dir):
+        os.makedirs(wheel_dir)
+
+    reqd_tag = get_priority_by_name(abi)
+    out_wheel = repair_wheel(wheel_path, abi, wheel_dir)
+
+    if out_wheel is not None:
```
ShaneHarvey, line 33: What if out_wheel is None? Should we error?

bindings/python/addtags.py
```diff
@@ -0,0 +1,49 @@
...
+        analyzed_tag = analyze_wheel_abi(out_wheel).overall_tag
+        if reqd_tag < get_priority_by_name(analyzed_tag):
+            print('Wheel is eligible for a higher priority tag. '
+                  'You requested %s but I have found this wheel is '
+                  'eligible for %s.' % (abi, analyzed_tag))
+            out_wheel = repair_wheel(wheel_path, analyzed_tag, wheel_dir)
```
ShaneHarvey, line 39: Can you explain this code?

bindings/python/addtags.py
prashantmital: auditwheel doesn't, so I would think not.

bindings/python/addtags.py
prashantmital: This happens when you build a manylinux1 wheel and attempt to assign it a manylinux2010 tag. Then auditwheel essentially tells you that you can support installation on a wider array of systems by simply using a more permissive ('higher priority') tag and then that's what it does.
See this file for a list of policies and associated priorities - https://github.com/pypa/auditwheel/blob/master/auditwheel/policy/policy.json - you will note that the most permissive policy, i.e. the policy with the lowest minimum requirements have highest priority.

ShaneHarvey (approved):

PR #31, opened 2021-04-27
BUMP 0.1.1

prashantmital:


ShaneHarvey (commented):
Do we need to update the changelog as well?

bindings/python/docs/source/conf.py
```diff
@@ -23,7 +23,7 @@
 html_show_sphinx = False
 
 # The full version, including alpha/beta/rc tags
-version = '0.2.0.dev0'
```
ShaneHarvey, line 26: Why do we need to duplicate the version here?

prashantmital: To display in the docs. In PyMongo we don't need to do this because the package can be installed and the version imported... I don't think that will work here.

ShaneHarvey: Can you add a comment here and in `version.py` to make sure these two stay in sync?

prashantmital: Actually, I remembered a way I handled this in another project... See the next commit.

ShaneHarvey (approved):

bindings/python/docs/source/conf.py
ShaneHarvey: Nice!

PR #36, opened 2021-11-19
PYTHON-3008 Test Python 3.10 support

juliusgeo:


ShaneHarvey:
We may need to upgrade to pyarrow 6.0 for 3.10 support. 6.0 is the first release to include wheels for 3.10: https://pypi.org/project/pyarrow/6.0.0/#files

bindings/python/requirements/build.txt
```diff
@@ -1,3 +1,3 @@
 setuptools>=47
 cython>=0.29
-pyarrow>=3,<3.1
\ No newline at end of file
+pyarrow>=3
```
ShaneHarvey, line 3: I suspect we need to update to `>=6,<6.1` because of the way we compile/link against pyarrow.

ShaneHarvey (changes requested):

ShaneHarvey (approved):

PR #40, opened 2021-12-15
PYTHON-2922 Add support for custom projections for find_arrow_all

blink1073:


bindings/python/pymongoarrow/api.py
```diff
@@ -103,7 +103,15 @@ def aggregate_arrow_all(collection, pipeline, *, schema, **kwargs):
...
-    pipeline.append({"$project": schema._get_projection()})
+    # Add a projection stage if none is given.
+    has_projection = False
+    for item in pipeline:
+        if '$project' in item:
+          has_projection = True
+          break
+    if not has_projection:
+        pipeline.append({"$project": schema._get_projection()})
+
```
ShaneHarvey, line 114: I think we should revert the aggregation change. Agg is different from find() in that an agg pipeline could have multiple intermediate `$project` stages in it:
```python
coll.aggregate([..., {"$project": ...}, ..., {"$project": ...}, ...])
```
I believe it's always ok to append our $project stage to the end. Even if it turns out our $project is redundant, the server should optimize it away anyway.

bindings/python/test/test_arrow.py
```diff
@@ -72,6 +72,21 @@ def test_find_simple(self):
...
+        projection = {'_id': True, 'data': {'$multiply': [2, '$data']}}
+        table = self.run_find({'_id': {'$gt': 2}},
+                              schema=self.schema,
+                              sort=[('_id', DESCENDING)],
+                              projection=projection)
+        self.assertEqual(table, expected)
+
+        find_cmd = self.cmd_listener.results['started'][-1]
+        self.assertEqual(find_cmd.command_name, 'find')
+        self.assertEqual(find_cmd.command['projection'], projection)
```
ShaneHarvey, line 88: Nice.

bindings/python/pymongoarrow/api.py
```diff
@@ -58,13 +58,13 @@ def find_arrow_all(collection, query, *, schema, **kwargs):
     context = PyMongoArrowContext.from_schema(
         schema, codec_options=collection.codec_options)
 
-    for opt in ('session', 'cursor_type', 'projection'):
+    for opt in ('session', 'cursor_type'):
```
ShaneHarvey, line 61: I believe with Python >=3.12 we _can_ support the session arg (added in https://jira.mongodb.org/browse/PYTHON-2389). Do you mind opening a new ticket to enable that feature here?

ShaneHarvey (changes requested):

bindings/python/pymongoarrow/api.py
blink1073: Done

ShaneHarvey: We don't need the extra new line now.

ShaneHarvey (changes requested):

ShaneHarvey (approved):

PR #41, opened 2021-12-16
PYTHON-3050 Add session support to find_raw_batches

blink1073:


bindings/python/test/test_arrow.py
```diff
@@ -87,6 +87,16 @@ def test_find_projection(self):
         self.assertEqual(find_cmd.command_name, 'find')
         self.assertEqual(find_cmd.command['projection'], projection)
 
+    def test_find_with_session(self):
+        with self.client.start_session() as session:
+            last_use = session._server_session.last_use
```
ShaneHarvey, line 92: `session._server_session` is private so we shouldn't use it unless we have to. Instead you can use the `session.operation_time` property:
```
self.assertIsNone(session.operation_time)
...
self.assertIsNotNone(session.operation_time)
```

ShaneHarvey (changes requested):

bindings/python/pymongoarrow/api.py
```diff
@@ -58,7 +58,7 @@ def find_arrow_all(collection, query, *, schema, **kwargs):
     context = PyMongoArrowContext.from_schema(
         schema, codec_options=collection.codec_options)
 
-    for opt in ('session', 'cursor_type'):
+    for opt in ('cursor_type'):
```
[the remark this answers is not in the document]
blink1073: Fixed

bindings/python/test/test_arrow.py
blink1073: Updated

ShaneHarvey (approved):

PR #42, opened 2021-12-17
PYTHON-2691 Add Windows support

blink1073:
WIP conda-forge recipe is [here](https://github.com/conda-forge/staged-recipes/pull/17302) (requires a release after merging this PR).


.github/workflows/test-python.yml
```diff
@@ -30,14 +30,14 @@ jobs:
           mongodb-replica-set: test-rs
       - name: Install libbson
         run: |
-          LIBBSON_INSTALL_DIR=$(pwd)/libbson ./build-libbson.sh
```
ShaneHarvey, line 33: Why rename LIBBSON_INSTALL_DIR to MONGO_LIBBSON_DIR? Asking because it makes this PR a little more difficult to review.

bindings/python/setup.py
```diff
@@ -10,9 +10,15 @@
...
 
-if platform not in ('linux', 'darwin'):
-    raise RuntimeError("Unsupported plaform {}".format(platform))
+# Find and copy the binary arrow files, unless
+# MONGO_NO_COPY_ARROW_LIB is set (for instance in a conda build).
+COPY_LIBARROW = not os.environ.get("MONGO_NO_COPY_LIBARROW", False)
+
+# Find and copy the binary libbson file, unless
+# MONGO_NO_COPY_LIBBSON is set (for instance in a conda build).
+COPY_LIBBSON = not os.environ.get("MONGO_NO_COPY_LIBBSON", False)
```
ShaneHarvey, line 21: Can you explain why we don't want to copy libbson (and libarrow) for conda builds? We should also record why we made this decision in the Jira ticket + commit message so we can refer back to it in the future.

bindings/python/setup.py
```diff
@@ -25,15 +31,46 @@ def query_pkgconfig(cmd):
...
+                    raise ValueError('We require a MONGO_LIBBSON_DIR with a compiled library on Windows')
             pc_path = os.path.join(
-                install_dir, libdir, 'pkgconfig', 'libbson-1.0.pc')
+                install_dir, lib_dir, 'pkgconfig', 'libbson-1.0.pc')
+
+    elif os.name == 'nt':
+        raise ValueError('We require a MONGO_LIBBSON_DIR with a compiled library on Windows')
+
+    if os.name == 'nt':
+        # We have either added the library file or raised an error, so return.
```
ShaneHarvey, line 72: Is "or raised an error" correct? If we raised an error we would never reach this line.

bindings/python/setup.py
```diff
@@ -25,15 +31,46 @@ def query_pkgconfig(cmd):
...
-            warnings.warn("Unable to locate {}".format('libbson-1.0.pc'))
+        # Handle the copy-able library file if applicable.
+        if COPY_LIBBSON:
+            if platform == "darwin":
+                lib_file = "libbson-1.0.0.dylib"
+            elif platform == "linux":
+                lib_file = "libbson-1.0.so.0"
+            else:  # windows
+                lib_file = 'bson-1.0.dll'
+            lib_dir = 'bin' if os.name == 'nt' else 'lib*'
```
ShaneHarvey, line 44: I suggest making `os.name == 'nt'` a variable to ease readability. I'm also more familiar with `sys.platform == 'win32'` which is what we use in other project to detect windows.

bindings/python/setup.py
```diff
@@ -25,15 +31,46 @@ def query_pkgconfig(cmd):
...
+
+        # Find the linkable library file, and explicity add it to the linker if on Windows.
+        lib_dirs = glob.glob(os.path.join(install_dir, "lib*"))
+        if len(lib_dirs) != 1:
+            warnings.warn(f"Unable to locate libbson in {install_dir}")
         else:
-            libdir = libdirs[0]
+            lib_dir = lib_dirs[0]
+            if os.name == 'nt':
+                lib_path = os.path.join(lib_dir, 'bson-1.0.lib').replace(os.sep, '/')
```
ShaneHarvey, line 58: What's up with the `.replace(os.sep, '/')`?

bindings/python/setup.py
```diff
@@ -25,15 +31,46 @@ def query_pkgconfig(cmd):
...
+                if os.path.exists(lib_path):
+                    module.extra_link_args = [lib_path]
+                    include_dir = os.path.join(install_dir, 'include', 'libbson-1.0').replace(os.sep, '/')
+                    module.include_dirs.append(include_dir)
+                else:
+                    raise ValueError('We require a MONGO_LIBBSON_DIR with a compiled library on Windows')
```
ShaneHarvey, line 64: This error happens when MONGO_LIBBSON_DIR is set but the file doesn't exists, should we update the error message?

bindings/python/setup.py
```diff
@@ -49,78 +86,84 @@ def append_libbson_flags(module):
...
 
-    module.include_dirs.append(np.get_include())
-    module.include_dirs.append(pa.get_include())
-    module.library_dirs.extend(pa.get_library_dirs())
+    if os.name == 'nt':
+        module.include_dirs.append(np.get_include())
+        module.include_dirs.append(pa.get_include())
+    else:
+        module.extra_compile_args.append("-isystem" + pa.get_include())
+        module.extra_compile_args.append("-isystem" + np.get_include())
```
ShaneHarvey, line 104: What's this for? Is this a replacement for:
```
    module.include_dirs.append(np.get_include())
    module.include_dirs.append(pa.get_include())
```

bindings/python/setup.py
```diff
@@ -49,78 +86,84 @@ def append_libbson_flags(module):
...
+    # You can use MONGO_LIBARROW_DIR to explicitly set the location of the 
     # arrow libraries (for instance in a conda build).
     # We first check for an unmodified path to the library,
     # then look for a library file with a version modifier, e.g. libarrow.600.dylib.
-    arrow_lib = os.environ.get('MONGO_ARROW_LIBDIR', pa.get_library_dirs()[0])
+    arrow_lib = os.environ.get('MONGO_LIBARROW_DIR', pa.get_library_dirs()[0])
     if platform == "darwin":
         exts = ['.dylib', '.*.dylib']
     elif platform == 'linux':
         exts = ['.so', '.so.*']
```
ShaneHarvey, line 122: Doesn't there need to be a windows branch here?

.github/workflows/test-python.yml
```diff
@@ -30,14 +30,14 @@ jobs:
           mongodb-replica-set: test-rs
```
ShaneHarvey, line 30: Don't we need to run the test suite on windows too?

ShaneHarvey (changes requested):

.github/workflows/test-python.yml
blink1073: Only for consistency with the arrow directory, I'll revert.

bindings/python/setup.py
blink1073: I'll update the docstring.  It is already documented in the JIRA ticket why shared libraries are handled differently on conda vs wheels.

bindings/python/setup.py
blink1073: Updating

bindings/python/setup.py
blink1073: Will do

bindings/python/setup.py
blink1073: Added a comment, it is so bash can parse it

bindings/python/setup.py
blink1073: Updating

bindings/python/setup.py
```diff
@@ -49,78 +86,84 @@ def append_libbson_flags(module):
...
+        module.include_dirs.append(pa.get_include())
+    else:
+        module.extra_compile_args.append("-isystem" + pa.get_include())
+        module.extra_compile_args.append("-isystem" + np.get_include())
+        # Arrow's manylinux{2010, 2014} binaries are built with gcc < 4.8 which predates CXX11 ABI
+        # - https://uwekorn.com/2019/09/15/how-we-build-apache-arrows-manylinux-wheels.html
+        # - https://arrow.apache.org/docs/python/extending.html#example
+        if "std=" not in os.environ.get("CXXFLAGS", ""):
+            module.extra_compile_args.append("-std=c++11")
+            module.extra_compile_args.append("-D_GLIBCXX_USE_CXX11_ABI=0")
```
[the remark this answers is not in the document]
blink1073: I was following what [snowflake-connector-python](https://github.com/snowflakedb/snowflake-connector-python/blob/6619dfc3389b6f9f6422af6ecc02921184fec97e/setup.py#L135) did

bindings/python/setup.py
blink1073: Yes, and also what `snowflake-connector-python` used

bindings/python/setup.py
blink1073: No, adding a comment

.github/workflows/test-python.yml
blink1073: Updating

blink1073: That was easy 😅

ShaneHarvey (approved):
LGTM!

PR #44, opened 2022-01-20
ARROW-12 Support ObjectId datatype

blink1073:


bindings/python/test/test_bson.py
```diff
@@ -18,12 +18,12 @@
...
 from pymongoarrow.lib import process_bson_stream
 from pymongoarrow.schema import Schema
-from pymongoarrow.types import int32, int64
+from pymongoarrow.types import int64, ObjectId
 
 
 class TestBsonToArrowConversionBase(TestCase):
     def setUp(self):
-        self.schema = Schema({'_id': int32(),
+        self.schema = Schema({'_id': ObjectId,
```
ShaneHarvey, line 26: Are we now loosing test coverage for int32?

bindings/python/pymongoarrow/types.py
```diff
@@ -64,8 +71,13 @@ def _get_internal_typemap(typemap):
     internal_typemap = {}
     for fname, ftype in typemap.items():
         for checker, internal_id in _TYPE_CHECKER_TO_INTERNAL_TYPE.items():
-            if checker(ftype):
-                internal_typemap[fname] = internal_id
-                continue
+            # Catch error where the pyarrow checkers are looking for an `id`
+            # attribute that might not exist on non-pyarrow types
+            # (like ObjectId).
```
ShaneHarvey, line 76: I don't follow this comment. Can you give an example?

bindings/python/test/test_bson.py
blink1073: All of the arrow types are checked in `test_pandas.py`

bindings/python/pymongoarrow/types.py
blink1073: Updated

ShaneHarvey (changes requested):
Existing changes LGTM! One last thing is to update the docs for the supported data types: https://github.com/mongodb-labs/mongo-arrow/blob/main/bindings/python/docs/source/supported_types.rst and the changelog.

ShaneHarvey (approved):

PR #46, opened 2022-01-26
ARROW-2 Add support for BSON string type

blink1073:


blink1073:
I added [concurrency handling](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-concurrency-to-cancel-any-in-progress-job-or-run) to prevent backlogs of queued jobs.

bindings/python/test/test_bson.py
```diff
@@ -103,8 +106,8 @@ def test_simple(self):
 
         schema = Schema({'_id': ObjectId,
                          'data': int64(),
-                         'title': pyarrow.string() })
+                         'valid': pyarrow.bool_() })
```
ShaneHarvey, line 109: Can we change this to a type that we never plan to support?

bindings/python/pymongoarrow/types.py
```diff
@@ -27,14 +27,16 @@ class _BsonArrowTypes(enum.Enum):
...
 
 
 _TYPE_NORMALIZER_FACTORY = {
     Int64: lambda _: int64(),
     float: lambda _: float64(),
     int: lambda _: int64(),
     datetime: lambda _: timestamp('ms'),     # TODO: add tzinfo support
-    ObjectId: lambda _: ObjectId
+    ObjectId: lambda _: ObjectId,
+    str: lambda: string()
```
ShaneHarvey, line 39: Suggest adding a trailing comma to make future diffs nicer :)

bindings/python/pymongoarrow/types.py
```diff
@@ -47,7 +49,8 @@ def _is_objectid(obj):
     _atypes.is_int64: _BsonArrowTypes.int64,
     _atypes.is_float64: _BsonArrowTypes.double,
     _atypes.is_timestamp: _BsonArrowTypes.datetime,
-    _is_objectid: _BsonArrowTypes.objectid
+    _is_objectid: _BsonArrowTypes.objectid,
+    _atypes.is_string: _BsonArrowTypes.string
```
ShaneHarvey, line 53: Suggest adding a trailing comma to make future diffs nicer :)

bindings/python/pymongoarrow/context.py
```diff
@@ -23,7 +23,8 @@
     _BsonArrowTypes.int64: Int64Builder,
     _BsonArrowTypes.double: DoubleBuilder,
     _BsonArrowTypes.datetime: DatetimeBuilder,
-    _BsonArrowTypes.objectid: ObjectIdBuilder
+    _BsonArrowTypes.objectid: ObjectIdBuilder,
+    _BsonArrowTypes.string: StringBuilder
```
ShaneHarvey, line 27: Suggest adding a trailing comma to make future diffs nicer :)

bindings/python/pymongoarrow/builders.pyi
```diff
@@ -19,6 +19,39 @@ cdef class _ArrayBuilderBase:
...
+    cpdef append_null(self):
+        self.builder.get().AppendNull()
+
+    def __len__(self):
+        return self.builder.get().length()
+
+    cpdef append(self, value):
+        if value is None or value is np.nan:
+            self.builder.get().AppendNull()
+        elif isinstance(value, (bytes, str)):
```
ShaneHarvey, line 40: Should we support coercing bytes to string? `bytes` is a separate type in BSON. Also are there any tests for this behavior?

ShaneHarvey (changes requested):

bindings/python/test/test_bson.py
blink1073: Done

bindings/python/pymongoarrow/types.py
blink1073: Done

bindings/python/pymongoarrow/types.py
blink1073: Done

bindings/python/pymongoarrow/context.py
blink1073: Done

bindings/python/pymongoarrow/builders.pyi
blink1073: When we call `bson_iter_utf8` it returns a `char *`, which gets converted by Cython to `bytes`.  We are guarded by using `value_t == BSON_TYPE_UTF8` in `bson.pyi`.   We'll use the `BinaryBuilder` for `BSON_TYPE_BINARY`.

ShaneHarvey: Hmm okay, so if we only ever call this method with bytes why should we check for None, np.nan, str, and other types here? Can this method be simplified to:
```python
cpdef append(self, value):
    self.builder.get().Append(tobytes(value))
```

blink1073: True, this logic was copied from other builders (which in turn was copied from `pyarrow.lib`).  I'll clean up the others as well to match how we're using them.

ShaneHarvey: I wonder if we should do this as a separate ticket. It would be interesting to know if there's a perf improvement for simplifying the builders.

ShaneHarvey (approved):
LGTM

bindings/python/pymongoarrow/builders.pyi
blink1073: Ah, I just saw this, do you want me to revert the last commit?  The builders aren't a public API, and the validation is done from `bson.pyi`, so I think it is a logical change.

ShaneHarvey: It's fine, I just didn't want to hold up this review since everything else already LGTM. I'l re review the new changes.

bindings/python/pymongoarrow/bson.pyi
```diff
@@ -75,6 +77,11 @@ def process_bson_stream(bson_stream, context):
                             builder.append(<bytes>(<uint8_t*>bson_iter_oid(&doc_iter))[:12])
                         else:
                             builder.append_null()
+                    elif ftype == t_string:
+                        if value_t == BSON_TYPE_UTF8:
+                            builder.append(bson_iter_utf8 (&doc_iter, &str_len))
```
ShaneHarvey, line 82: How does `tobytes` deal with nulls in the value? IIUC bson_iter_utf8 returns a pointer to the C byte string but it looks like we never use str_len.

blink1073: It is a null-terminated string, we have to provide a `str_len` to match the API, but we don't use it.

blink1073: Actually, I see that UTF-8 can have NUL characters embedded, so we do need to handle that case.

ShaneHarvey (approved):
LGTM

gaurav-basesolve:
is this gonna be available in next release? And when ?

blink1073:
Hi @gaurav-basesolve, yes, this will be in the next release.  We're aiming to release tomorrow.

PR #49, opened 2022-01-28
ARROW-21 Test PyMongoArrow documentation build in CI

blink1073:
Also addresses ARROW-41

### DIFF
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -65,27 +65,3 @@ jobs:
         run: |
           pip install check-manifest
           check-manifest -v
-
-  docs:
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: [3.9]
-    name: Documentation
-    steps:
-      - uses: actions/checkout@v2
-      - name: Setup conda
-        uses: conda-incubator/setup-miniconda@v2
-        with:
-          auto-update-conda: true
-          activate-environment: mongo_arrow_documentation
-          miniforge-variant: Mambaforge
-          environment-file: bindings/python/docs/environment.yml
-          python-version: ${{ matrix.python-version }}
-      - name: Build docs
-        shell: bash -l {0}
-        working-directory: ./bindings/python/docs
-        run: |
-          set -eux
-          pip install ..
-          make html SPHINXOPTS="-W"
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -71,7 +71,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Base Setup
         uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
-     - name: Setup conda
+      - name: Setup conda
         uses: conda-incubator/setup-miniconda@v2
         with:
           auto-update-conda: true
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -84,5 +84,4 @@ jobs:
         working-directory: ./bindings/python/docs
         run: |
           set -eux
-          pip install ..
           make html SPHINXOPTS="-W"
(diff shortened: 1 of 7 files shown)


.github/workflows/test-python.yml
```diff
@@ -25,10 +25,8 @@ jobs:
     name: CPython ${{ matrix.python-version }}-${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
-      - name: Setup Python
-        uses: actions/setup-python@v2
-        with:
-          python-version: ${{ matrix.python-version }}
+      - name: Base Setup
+        uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
```
blink1073, line 29: This action takes care of installing python and setting up caching for pip and conda

bindings/python/setup.py
```diff
@@ -38,43 +38,52 @@ def query_pkgconfig(cmd):
...
-
-        # Find the linkable library file, and explicity add it to the linker if on Windows.
-        lib_dirs = glob.glob(os.path.join(install_dir, "lib*"))
-        if len(lib_dirs) != 1:
-            warnings.warn(f"Unable to locate libbson in {install_dir}")
+    dynamic_lib_dir = 'bin' if IS_WIN else 'lib*'
+    static_lib_dir = 'lib*'
+
+    # If no install dir is given, try looking for an installed python
+    # package (e.g. libbson from conda-forge).
```
blink1073, line 45: This will also be useful if we create a libbson package and publish to PyPI.

blink1073:
Moving back to draft so I can create the appropriate RTD config file.

ShaneHarvey (commented):
There are surprisingly a lot of changes to setup.py here. Can you explain?

blink1073:
> There are surprisingly a lot of changes to setup.py here. Can you explain?

I was refactoring to support installing from the `libbson` `conda` package without requiring an environment variable (since we can't set it on RTD).  I also wanted to preserve the ability to use the system `libbson` (I have the `homebrew` version installed locally).  I noticed that the `IS_WIN` vs Posix handling branch were getting muddled, so I separated it out into different functions, which allowed me to clean it up a lot.

ShaneHarvey:
Interesting. Is supporting libbson conda needed? It seems like we could just run `./build-libbson.sh` like usual before building the docs.

blink1073:
> It seems like we could just run ./build-libbson.sh like usual before building the docs.

ReadTheDocs does not yet support running scripts prior to `pip install .`.  However, I found the solution: if you install `pkg-config` in the `conda` environment, then that version of `pkg-config` looks in the conda environment for its libraries, and it works with no changes to `setup.py`!

.readthedocs.yaml
```diff
@@ -0,0 +1,10 @@
+version: 2
+sphinx:
+  configuration: ./bindings/python/docs/source/conf.py
+conda:
+  environment: ./bindings/python/docs/environment.yml
+python:
+  install:
+    # install the package itself
+    - method: pip
+      path: ./bindings/python
```
ShaneHarvey, line 10: Does this build properly properly on readthedocs now that the workarounds are removed? If so do we need jupyterlab at all?

blink1073: We're only using JupyterLab Maintainer-Tools to simplify the Actions logic, it saves us from having to keep all those things up to date and in sync.

blink1073: The PR build works!

blink1073: Removed usage of the jupyterlab action

bindings/python/docs/environment.yml
```diff
@@ -0,0 +1,8 @@
+name: mongo_arrow_documentation
+channels:
+  - conda-forge
+dependencies:
+- python=3.9
+- sphinx
+- libbson
```
ShaneHarvey, line 7: no longer needed?

ShaneHarvey (approved):
LGTM with one minor comment

bindings/python/docs/environment.yml
blink1073: We are still using it, as the "system" `libbson`

ShaneHarvey: Thanks!

PR #50, opened 2022-02-01
ARROW-54 Make ObjectId an Extension Type

blink1073:


ShaneHarvey (approved):
LGTM!

ShaneHarvey (changes requested):
Whoops, we need to update the supported types docs page now too.

bindings/python/pymongoarrow/types.py
```diff
@@ -30,20 +31,35 @@ class _BsonArrowTypes(enum.Enum):
     string = 6
 
 
+# Custom Extension Types.
+
+class ObjectIdType(PyExtensionType):
+    _type_marker = _BsonArrowTypes.objectid
```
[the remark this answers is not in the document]
blink1073: Yeah, I read Jesse's blog post about it recently.  :)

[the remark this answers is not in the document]
ShaneHarvey: Oh I thought this was a PyExtensionType thing.

blink1073:
> Whoops, we need to update the supported types docs page now too.

Updated

bindings/python/pymongoarrow/types.py
```diff
@@ -30,20 +31,35 @@ class _BsonArrowTypes(enum.Enum):
...
+
+    def __init__(self):
+        PyExtensionType.__init__(self, binary(12))
```
ShaneHarvey, line 40: Should we use super() instead?

bindings/python/pymongoarrow/types.py
```diff
@@ -30,20 +31,35 @@ class _BsonArrowTypes(enum.Enum):
...
+
+    def __reduce__(self):
+        return ObjectIdType, ()
```
ShaneHarvey, line 43: Any particular reason you chose to use reduce here https://docs.python.org/3/library/pickle.html#object.__setstate__

> Although powerful, implementing __reduce__() directly in your classes is error prone. For this reason, class designers should use the high-level interface (i.e., __getnewargs_ex__(), __getstate__() and __setstate__()) whenever possible. We will show, however, cases where using __reduce__() is the only option or leads to more efficient pickling or both.

Also should we test that this actually works?

bindings/python/docs/source/supported_types.rst
```diff
@@ -18,7 +18,7 @@ Support for additional types will be added in subsequent releases.
    * - String
      - :class:`py.str`, an instance of :class:`pyarrow.string`
    * - ObjectId
-     - :class:`py.bytes`, :class:`bson.ObjectId`, an instance of :class:`pyarrow.FixedSizeBinaryScalar`
+     - :class:`py.bytes`, :class:`bson.ObjectId`, an instance of :class:`pymongoarrow.types.ObjectIdType`, an instance of :class:`pyarrow.FixedSizeBinaryScalar`
    * - 64-bit binary floating point
      - :class:`py.float`, an instance of :meth:`pyarrow.float64`
    * - 32-bit integer
```
ShaneHarvey, line 24: Does the schema example below also need ObjectId -> ObjectIdType()?

blink1073: Either one works, thanks to `_normalize_typeid`

bindings/python/pymongoarrow/types.py
blink1073: I pulled this pattern from the arrow [docs](https://arrow.apache.org/docs/python/extending_types.html#defining-extension-types-user-defined-types).  I'll use the advanced `__arrow_ext_serialize__` method instead.    Agreed we should test it either way.

bindings/python/pymongoarrow/types.py
blink1073: Will do

bindings/python/pymongoarrow/types.py
ShaneHarvey: Oh in that case can you just add a comment referencing that docs page? I don't think we need to use the longer `__arrow_ext_serialize__` methods yet since we don't need their functionality.

blink1073: Done and tested

ShaneHarvey (approved):
LGTM!

PR #58, opened 2022-02-28
ARROW-58 Add PyArrow 7.0 Support

blink1073:


ShaneHarvey (changes requested):
Let's update the install guide too:
```
bindings/python/docs/source/installation.rst:14:PyMongoArrow is currently compatible with CPython 3.6, 3.7, 3.8 and 3.9.
```

ShaneHarvey (approved):

PR #68, opened 2022-04-05
ARROW-72 Run performance tests on prototype and evaluate our approach

juliusgeo:


bindings/python/benchmark.py
```diff
@@ -143,6 +152,17 @@ def to_arrow(use_large):
     find_arrow_all(c, {}, schema=schema)
 
 
+@bench("arrow-roundtrip")
+def roundtrip_arrow(use_large):
```
ShaneHarvey, line 156: "roundtrip" to me implies read + write while this benchmark only cares about the write. Let's rename to "conventional":
```
@bench("insert-arrow")
def insert_arrow(use_large):
...
@bench("insert-arrow-conventional")
def insert_arrow_conventional(use_large):
```

ShaneHarvey (changes requested):

bindings/python/benchmark.py
```diff
@@ -143,6 +152,17 @@ def to_arrow(use_large):
     find_arrow_all(c, {}, schema=schema)
 
 
+@bench("insert_arrow")
+def insert_arrow(use_large):
+    write(db[collection_names[use_large]], arrow_tables[use_large])
+
+
+@bench("insert_naive")
+def insert_naive(use_large):
```
ShaneHarvey, line 161: naive->conventional (the term the rest of the file uses).

bindings/python/pymongoarrow/api.py
```diff
@@ -279,23 +279,24 @@ def write(collection, tabular):
...
             enc_tab = RawBSONDocument(
                 encode(next(tabular_gen), codec_options=collection.codec_options)
             )
             cur_batch.append(enc_tab)
-            cur_size += len(enc_tab)
+            cur_size += len(enc_tab.raw)
             i += 1
         try:
-            collection.insert_many(cur_batch)
+            collection.insert_many(cur_batch, bypass_document_validation=True)
```
ShaneHarvey, line 299: Why add bypass_document_validation=True?

ShaneHarvey (changes requested):

bindings/python/pymongoarrow/api.py
juliusgeo: That was in error. Removed.

ShaneHarvey (approved):
LGTM!

PR #71, opened 2022-04-09
ARROW-74 Add support and testing for NumPy

juliusgeo:
Benchmark results:
```
(python3.8) ➜  python git:(ARROW-74) ✗ python benchmark.py

100000 small docs, 40 bytes each with 3 keys
1000 large docs, 153k each with 2600 keys

                    BENCH:   SMALL   LARGE
  conventional-to-ndarray:    0.20    1.13 
    pymongoarrow-to-numpy:    0.29    1.69 
   conventional-to-pandas:    0.26    1.74 
   pymongoarrow-to-pandas:    0.28    1.67 
    pymongoarrow-to-arrow:    0.28    1.67 
             insert_arrow:    1.79    4.78 
      insert_conventional:    1.86    4.86 
            insert_pandas:    2.24    5.04 
             insert_numpy:    2.15    4.66 
```

### DIFF
--- a/bindings/python/pymongoarrow/api.py
+++ b/bindings/python/pymongoarrow/api.py
@@ -13,8 +13,10 @@
 # limitations under the License.
 import warnings
 
+import numpy as np
 from bson import encode
 from bson.raw_bson import RawBSONDocument
+from numpy import ndarray
 from pandas import DataFrame
 from pyarrow import Schema as ArrowSchema
 from pyarrow import Table
@@ -25,7 +27,7 @@ from pymongoarrow.errors import ArrowWriteError
 from pymongoarrow.lib import process_bson_stream
 from pymongoarrow.result import ArrowWriteResult
 from pymongoarrow.schema import Schema
-from pymongoarrow.types import _validate_schema
+from pymongoarrow.types import _validate_schema, get_numpy_type
 
 __all__ = [
     "aggregate_arrow_all",
@@ -187,7 +189,11 @@ def _arrow_to_numpy(arrow_table, schema):
     """
     container = {}
     for fname in schema:
-        container[fname] = arrow_table[fname].to_numpy()
+        dtype = get_numpy_type(schema.typemap[fname])
+        if dtype == np.str_:
+            container[fname] = arrow_table[fname].to_pandas().to_numpy(dtype=dtype)
+        else:
+            container[fname] = arrow_table[fname].to_numpy()
     return container
 
 
@@ -264,8 +270,15 @@ def _tabular_generator(tabular):
             for row in i.to_pylist():
                 yield row
     elif isinstance(tabular, DataFrame):
-        for i in tabular.to_dict("records"):
-            yield i
+        for row in tabular.to_dict("records"):
+            yield row
+    elif isinstance(tabular, dict):
+        iter_dict = {k: iter(v.tolist()) for k, v in tabular.items()}
+        while True:
+            try:
+                yield {k: next(i) for k, i in iter_dict.items()}
+            except StopIteration:
+                break
 
 
 def write(collection, tabular):
@@ -279,17 +292,23 @@ def write(collection, tabular):
     :Returns:
       An instance of :class:`result.ArrowWriteResult`.
     """
-
+    tab_size = len(tabular)
     if isinstance(tabular, Table):
         _validate_schema(tabular.schema.types)
     elif isinstance(tabular, DataFrame):
         _validate_schema(ArrowSchema.from_pandas(tabular).types)
+    elif (
+        isinstance(tabular, dict)
+        and len(tabular.values()) >= 1
+        and isinstance(list(tabular.values())[0], ndarray)
+    ):
+        _validate_schema([i.dtype for i in tabular.values()])
+        tab_size = len(list(tabular.values())[0])
     cur_offset = 0
     results = {
         "insertedCount": 0,
     }
     tabular_gen = _tabular_generator(tabular)
-    tab_size = len(tabular)
     while cur_offset < tab_size:
         cur_size = 0
         cur_batch = []
--- a/bindings/python/pymongoarrow/api.py
+++ b/bindings/python/pymongoarrow/api.py
@@ -292,6 +292,10 @@ def write(collection, tabular):
     :Returns:
       An instance of :class:`result.ArrowWriteResult`.
     """
+    cur_offset = 0
+    results = {
+        "insertedCount": 0,
+    }
     tab_size = len(tabular)
     if isinstance(tabular, Table):
         _validate_schema(tabular.schema.types)
@@ -304,10 +308,9 @@ def write(collection, tabular):
     ):
         _validate_schema([i.dtype for i in tabular.values()])
         tab_size = len(next(iter(tabular.values())))
-    cur_offset = 0
-    results = {
-        "insertedCount": 0,
-    }
+    else:
+        raise ArrowWriteError(results)
+
     tabular_gen = _tabular_generator(tabular)
     while cur_offset < tab_size:
         cur_size = 0
--- a/bindings/python/pymongoarrow/api.py
+++ b/bindings/python/pymongoarrow/api.py
@@ -273,10 +273,10 @@ def _tabular_generator(tabular):
         for row in tabular.to_dict("records"):
             yield row
     elif isinstance(tabular, dict):
-        iter_dict = {k: iter(v.tolist()) for k, v in tabular.items()}
+        iter_dict = {k: np.nditer(v) for k, v in tabular.items()}
         while True:
             try:
-                yield {k: next(i) for k, i in iter_dict.items()}
+                yield {k: next(i).item() for k, i in iter_dict.items()}
             except StopIteration:
                 break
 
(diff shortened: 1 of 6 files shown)


bindings/python/pymongoarrow/api.py
```diff
@@ -264,8 +270,15 @@ def _tabular_generator(tabular):
             for row in i.to_pylist():
                 yield row
     elif isinstance(tabular, DataFrame):
-        for i in tabular.to_dict("records"):
-            yield i
+        for row in tabular.to_dict("records"):
+            yield row
+    elif isinstance(tabular, dict):
+        iter_dict = {k: iter(v.tolist()) for k, v in tabular.items()}
```
blink1073, line 276: Does this put the whole array into memory?  Maybe look into https://numpy.org/doc/stable/reference/generated/numpy.nditer.html#numpy.nditer

juliusgeo: Good catch. Fixed.

bindings/python/pymongoarrow/api.py
```diff
@@ -264,8 +270,15 @@ def _tabular_generator(tabular):
             for row in i.to_pylist():
                 yield row
     elif isinstance(tabular, DataFrame):
-        for i in tabular.to_dict("records"):
-            yield i
+        for row in tabular.to_dict("records"):
+            yield row
+    elif isinstance(tabular, dict):
+        iter_dict = {k: np.nditer(v) for k, v in tabular.items()}
```
blink1073, line 276: Can we not iterate over this directly rather than explicitly calling next?

juliusgeo: That seems unintuitive to me. Would that not just be calling next, but in a different and less clear way? We are going to have to call  ``ndarray.item()`` anyways, so I don't think that there is any way to avoid the dictionary comprehension. I also cannot figure out how to implement it your way. With a for loop it calls next on the container, but not on the individual elements thereof. So we basically are forced to call next ourselves, if I understand it correctly.

blink1073: Fair enough

blink1073 (approved):
Thanks!

bindings/python/pymongoarrow/api.py
```diff
@@ -279,17 +292,23 @@ def write(collection, tabular):
...
-
+    tab_size = len(tabular)
     if isinstance(tabular, Table):
         _validate_schema(tabular.schema.types)
     elif isinstance(tabular, DataFrame):
         _validate_schema(ArrowSchema.from_pandas(tabular).types)
+    elif (
+        isinstance(tabular, dict)
+        and len(tabular.values()) >= 1
+        and isinstance(list(tabular.values())[0], ndarray)
```
ShaneHarvey, line 303: Do we need to check `tabular.values()` in this if statement? What happens if `tabular` is a dict with different format?

bindings/python/pymongoarrow/api.py
```diff
@@ -279,17 +292,23 @@ def write(collection, tabular):
...
+    ):
+        _validate_schema([i.dtype for i in tabular.values()])
+        tab_size = len(list(tabular.values())[0])
```
ShaneHarvey, line 306: The list() call here is wasteful since we only need the first value. How about `len(next(iter(tabular.values())))`?

bindings/python/pymongoarrow/types.py
```diff
@@ -105,6 +125,11 @@ def _get_internal_typemap(typemap):
 
 
 def _in_type_map(t):
+    if isinstance(t, np.dtype):
+        try:
+            t = pa.from_numpy_dtype(t)
+        except pa.lib.ArrowNotImplementedError:
+            return False
```
ShaneHarvey, line 132: What prompted these changes?

bindings/python/test/test_numpy.py
```diff
@@ -45,13 +48,14 @@ def setUp(self):
         self.cmd_listener.reset()
         self.getmore_listener.reset()
 
-    def assert_numpy_equal(self, actual, expected):
+    def assert_numpy_equal(self, actual, expected, schema=None):
         self.assertIsInstance(actual, dict)
-        for field in self.schema:
+        for field in schema or self.schema:
+            # print(set(actual)-set(e))
```
ShaneHarvey, line 54: Unneeded?

bindings/python/test/test_numpy.py
```diff
@@ -45,13 +48,14 @@ def setUp(self):
         self.cmd_listener.reset()
         self.getmore_listener.reset()
 
-    def assert_numpy_equal(self, actual, expected):
+    def assert_numpy_equal(self, actual, expected, schema=None):
         self.assertIsInstance(actual, dict)
-        for field in self.schema:
+        for field in schema or self.schema:
```
ShaneHarvey, line 53: Do we need to pass the schema here at all? What about just `for field in expected:`?

bindings/python/test/test_numpy.py
```diff
@@ -91,3 +95,84 @@ def test_aggregate_simple(self):
...
+        with self.assertRaises(ArrowWriteError):
+            try:
+                self.round_trip(data, Schema({"_id": int32(), "data": int64()}))
+            except ArrowWriteError as awe:
+                self.assertEqual(
+                    10001, awe.details["writeErrors"][0]["index"], awe.details["nInserted"]
+                )
+                raise awe
+
+    def test_write_schema_validation(self):
```
ShaneHarvey, line 128: I can't help but feel like all this "schema" management is painful. Should a schema even be required for reading? It should be fairly straightforward to auto-discover the schema based on the fields and types in the first document. Not really an issue in this PR itself but something to think about. Do we already have a ticket for this auto-discover schema idea?

bindings/python/pymongoarrow/api.py
```diff
@@ -187,7 +189,11 @@ def _arrow_to_numpy(arrow_table, schema):
     """
     container = {}
     for fname in schema:
-        container[fname] = arrow_table[fname].to_numpy()
+        dtype = get_numpy_type(schema.typemap[fname])
+        if dtype == np.str_:
+            container[fname] = arrow_table[fname].to_pandas().to_numpy(dtype=dtype)
+        else:
+            container[fname] = arrow_table[fname].to_numpy()
```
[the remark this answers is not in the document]
ShaneHarvey: Interesting, why didn't we notice this before? And why do we only use to_pandas() for np.str_? What about other types that might have nulls?

bindings/python/pymongoarrow/api.py
ShaneHarvey: The current code is still problematic, for example what happens when you call:
```
write({'foo': 1})
```

bindings/python/pymongoarrow/types.py
ShaneHarvey: Gotcha makes sense to me.

ShaneHarvey (changes requested):

bindings/python/pymongoarrow/api.py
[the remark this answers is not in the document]
ShaneHarvey: The docs say to_numpy cannot deal with nulls: 
> By default, tries to return a view of this array. This is only supported for primitive arrays with the same memory layout as NumPy (i.e. integers, floating point, ..) and without any nulls.

https://arrow.apache.org/docs/python/generated/pyarrow.Int32Array.html?highlight=to_numpy#pyarrow.Int32Array.to_numpy

Are the docs incorrect? What happens in this case?

Could you also open a new ticket for testing every supported data type in find_arrow/pandas/numpy?

bindings/python/pymongoarrow/api.py
```diff
@@ -279,17 +292,26 @@ def write(collection, tabular):
...
+        _validate_schema(ArrowSchema.from_pandas(tabular).types)
+    elif (
+        isinstance(tabular, dict)
+        and len(tabular.values()) >= 1
+        and all([isinstance(i, ndarray) for i in tabular.values()])
+    ):
+        _validate_schema([i.dtype for i in tabular.values()])
+        tab_size = len(next(iter(tabular.values())))
+    else:
+        raise ArrowWriteError(results)
```
ShaneHarvey, line 312: This should be a ValueError with a helpful error message explaining that tabular is the wrong type.

ShaneHarvey (changes requested):

bindings/python/pymongoarrow/api.py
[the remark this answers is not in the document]
juliusgeo: The docs are wrong. In ``test_find_simple`` we utilize an array with a None and a np.nan value. Both of them work correctly (meaning the null values are converted to np.nan or None). And I will open that ticket.

bindings/python/pymongoarrow/api.py
```diff
@@ -279,17 +292,30 @@ def write(collection, tabular):
...
+    elif (
+        isinstance(tabular, dict)
+        and len(tabular.values()) >= 1
+        and all([isinstance(i, ndarray) for i in tabular.values()])
+    ):
+        _validate_schema([i.dtype for i in tabular.values()])
+        tab_size = len(next(iter(tabular.values())))
+    else:
+        raise ValueError(
+            f"You passed an invalid tabular data object of type {type(tabular)} \n"
```
ShaneHarvey, line 313: "You passed an" is odd phrasing for an error message. Let's say:
`invalid tabular argument of type....`

ShaneHarvey (changes requested):

ShaneHarvey (approved):
LGTM!

PR #73, opened 2022-04-13
ARROW-71 Add test for writing from a Parquet file

juliusgeo:


ShaneHarvey (changes requested):
Instead of duplicating the whole test_arrow file, let's just add one test to TestArrowApiMixin which uses `pq.read_table`.

ShaneHarvey (approved):

blink1073 (approved):
👍🏼

PR #75, opened 2022-04-20
ARROW-91 Include tests in the sdist

blink1073:


.github/workflows/test-python.yml
```diff
@@ -47,7 +47,7 @@ jobs:
         if: ${{ startsWith(runner.os, 'Linux') }}
         uses: supercharge/mongodb-github-action@1.3.0
         with:
-          mongodb-version: 4.4
```
blink1073, line 50: 4.4 was no longer available

blink1073: Actually, it may have been spurious

blink1073: Yeah, it was a temporary 503 error

PR #76, opened 2022-04-20
ARROW-79 Handling PyMongo errors in PyMongoArrow Writing Support

juliusgeo:


bindings/python/pymongoarrow/errors.py
```diff
@@ -46,5 +46,10 @@ def details(self):
             'nRemoved': ...,
             'upserted': [...]
         }
+        Details will have the following more general format for other PyMongo errors:
+        {
+            'cause': ..., (a PyMongoError subclass object)
```
blink1073, line 51:
```suggestion
            'cause': ..., (a :class:`~pymongo.errors.PyMongoError` subclass object)
```

bindings/python/pymongoarrow/api.py
```diff
@@ -336,6 +337,8 @@ def write(collection, tabular):
             collection.insert_many(cur_batch)
         except BulkWriteError as bwe:
             raise ArrowWriteError(_transform_bwe(dict(bwe.details), cur_offset)) from bwe
+        except pymongo.errors.PyMongoError as pme:
+            raise ArrowWriteError({"cause": pme, "index": cur_offset}) from pme
```
ShaneHarvey, line 341: This makes for a somewhat unwieldy api since we raise ArrowWriteError in two different forms. Is there any way to unify the two apis?

Also I think we can get rid of "cause" since users can already access `exc. __cause__ ` set by the `raise ArrowWriteError(...) from pme`. See:
https://docs.python.org/3/tutorial/errors.html#exception-chaining  and https://docs.python.org/3/library/exceptions.html#exception-context for more info about chaining.

ShaneHarvey (changes requested):

bindings/python/pymongoarrow/api.py
juliusgeo: If we can get rid of cause, then we can raise an ArrowWriteError of the same form as the bulk write error (ie. with ``"writeErrors": {"errmsg": exc._message, "index":cur_offset}`` this form)

bindings/python/pymongoarrow/api.py
```diff
@@ -336,6 +337,10 @@ def write(collection, tabular):
             collection.insert_many(cur_batch)
         except BulkWriteError as bwe:
             raise ArrowWriteError(_transform_bwe(dict(bwe.details), cur_offset)) from bwe
+        except pymongo.errors.PyMongoError as pme:
+            raise ArrowWriteError(
+                {"writeErrors": [{"errmsg": pme._message, "index": cur_offset}]}
```
ShaneHarvey, line 342: `_message` is private, use `str(pme)` instead.

bindings/python/pymongoarrow/api.py
ShaneHarvey: Great!

bindings/python/pymongoarrow/api.py
```diff
@@ -336,6 +337,10 @@ def write(collection, tabular):
             collection.insert_many(cur_batch)
         except BulkWriteError as bwe:
             raise ArrowWriteError(_transform_bwe(dict(bwe.details), cur_offset)) from bwe
+        except pymongo.errors.PyMongoError as pme:
+            raise ArrowWriteError(
```
ShaneHarvey, line 341: Could we include `nInserted` and `writeConcernErrors` as well to match the BulkWriteError error case?

bindings/python/pymongoarrow/api.py
```diff
@@ -336,6 +337,10 @@ def write(collection, tabular):
             collection.insert_many(cur_batch)
         except BulkWriteError as bwe:
             raise ArrowWriteError(_transform_bwe(dict(bwe.details), cur_offset)) from bwe
```
ShaneHarvey, line 339: A few comments for consistency:

1. _transform_bwe should only include writeErrors+writeConcernErrors+nInserted and not the other fields because write() only runs insert.
2. This is unrelated to the current PR but can we rename results["insertedCount"] to just inserted_count = 0? Then we can have ArrowWriteResult accept a named parameter inserted_count. Is there any utility for making "results" a dictionary?

ShaneHarvey (changes requested):

bindings/python/pymongoarrow/api.py
juliusgeo: 1. Answered below.
2. I think the primary utility of making results a dictionary is that there is one consistent api for accessing all of the elements. You can also do things like iterate over all the fields more easily (like if you were trying to compare two results).

bindings/python/pymongoarrow/api.py
juliusgeo: I'm confused why you're in favor of adding what are essentially null fields here, but in the instance above you are not in favor of keeping null fields that we are going to make use of in the future.

bindings/python/pymongoarrow/api.py
juliusgeo: Done.

bindings/python/pymongoarrow/api.py
ShaneHarvey: I think we should add writeConcernErrors for consistency with the other error case. For `nInserted` it is not null, it's very possible to insert a batch of documents and then run into a non-BulkWriteError on the next batch. So adding `nInserted` is actually required.

bindings/python/pymongoarrow/api.py
ShaneHarvey: Fair enough.

bindings/python/pymongoarrow/errors.py
```diff
@@ -33,18 +33,14 @@ def __init__(self, details):
     def details(self):
         """Details for the error.
 
-        It is a dictionary of key-value pairs giving diagnostic information about what went wrong. To see the entire dictionary simply use `print(awe.details)`.
```
blink1073, line 36: Perhaps add a note that if the error was triggered by a `PyMongoError` then that error will be available on the `__cause__` attribute?  I wasn't aware of that feature.

ShaneHarvey (approved):

PR #83, opened 2022-07-19
ARROW-94 Test for nulls

bcwarner:
Adds unit tests for null value behaviors for pyarrow, numpy, and pandas. Leaves underlying behavior intact.

### DIFF
--- a/bindings/python/test/test_arrow.py
+++ b/bindings/python/test/test_arrow.py
@@ -426,8 +426,7 @@ class TestNulls(unittest.TestCase):
     def test_int_handling(self):
         # Default integral types
         int_schema = Schema({"_id": ObjectIdType(), "int64": int64()})
-        int64s = [(i if (i % 2 == 0) else None) for i in range(len(
-            self.oids))]
+        int64s = [(i if (i % 2 == 0) else None) for i in range(len(self.oids))]
         self.coll.insert_many(
             [{"_id": self.oids[i], "int64": int64s[i]} for i in range(len(self.oids))]
         )
@@ -443,21 +442,18 @@ class TestNulls(unittest.TestCase):
 
     def test_other_handling(self, na_safe=True, dtype="O"):
         # generating fn, arrow type, numpy dtype
-        other_pairs = [(str, string(), "O"),
-                       (int, int32(), ">i4"),
-                       (float, float64(), "d"),
-                       (lambda x: datetime.datetime(x + 1, 1, 1),
-                        timestamp("ms"),
-                        "datetime64[ms]"),
-                       (lambda _: ObjectId(), ObjectIdType(),
-                        "O"),
-                       (lambda x: Decimal128(str(x)), Decimal128StringType(),
-                        "O")]
+        other_pairs = [
+            (str, string(), "O"),
+            (int, int32(), ">i4"),
+            (float, float64(), "d"),
+            (lambda x: datetime.datetime(x + 1, 1, 1), timestamp("ms"), "datetime64[ms]"),
+            (lambda _: ObjectId(), ObjectIdType(), "O"),
+            (lambda x: Decimal128(str(x)), Decimal128StringType(), "O"),
+        ]
 
         for gen, atype, dtype in other_pairs:
             other_schema = Schema({"_id": ObjectIdType(), "other": atype})
-            others = [gen(i) if (i % 2 == 0) else None for i in range(len(
-                self.oids))]
+            others = [gen(i) if (i % 2 == 0) else None for i in range(len(self.oids))]
 
             self.setUp()
 
--- a/bindings/python/test/test_arrow.py
+++ b/bindings/python/test/test_arrow.py
@@ -11,18 +11,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import datetime
 import os
 import unittest
 import unittest.mock as mock
 from test import client_context
 from test.utils import AllowListEventListener
 
-import numpy as np
 import pyarrow
 import pymongo
 from bson import Decimal128, ObjectId
-from pandas import isna
 from pyarrow import Table, binary, bool_, decimal256, float64, int32, int64
 from pyarrow import schema as ArrowSchema
 from pyarrow import string, timestamp
@@ -37,6 +34,8 @@ from pymongoarrow.types import (
     Decimal128StringType,
     ObjectIdType,
 )
+
+from .utils import TestNullsBase
 
 
 class TestArrowApiMixin:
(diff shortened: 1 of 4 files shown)


bindings/python/test/test_arrow.py
```diff
@@ -384,3 +387,84 @@ def test_find_decimal128(self):
...
+        cls.cmd_listener = AllowListEventListener("find", "aggregate")
+        cls.getmore_listener = AllowListEventListener("getMore")
+        cls.client = client_context.get_client(
+            event_listeners=[cls.getmore_listener, cls.cmd_listener]
+        )
+
+        cls.oids = [ObjectId() for _ in range(4)]
+
+        # Default integral types
+        cls.int_schema = Schema({"_id": ObjectIdType(), "int64": int64()})
```
juliusgeo, line 406: These arrays and schemas are only used in their respective test, could they be moved to local variables?

bindings/python/test/test_arrow.py
```diff
@@ -384,3 +387,84 @@ def test_find_decimal128(self):
...
+        )
+
+        table = self.find_fn(self.coll, {}, schema=self.int_schema)
+
+        # Resulting datatype should be float64 according to the spec for numpy
+        # and pandas.
+        self.assertType(table["int64"], "float64", int64())
+
+        # Does it contain NAs where we expect?
+        self.assertTrue(np.all(np.equal(isna(self.int64s), isna(table["int64"]))))
```
juliusgeo, line 445: What about a 32 bit integer?

bindings/python/test/test_arrow.py
```diff
@@ -384,3 +389,125 @@ def test_find_decimal128(self):
...
+        )
+
+        cls.oids = [ObjectId() for _ in range(4)]
+        cls.coll = cls.client.pymongoarrow_test.get_collection(
+            "test", write_concern=WriteConcern(w="majority")
+        )
+
+        cls.find_fn = lambda s, a1, a2, schema=None: find_fn(a1, a2, schema=schema)
+        cls.equal_fn = equal_fn
+        cls.table_from_dict = lambda s, d: table_from_dict(d)
```
ShaneHarvey, line 417: Rather than injecting these methods via setUpClass, what do you think about using methods, like:
```python
class TestNulls(unittest.TestCase):
...
    def find(self, ...):
        return find_arrow_all(...)
```

Then the other subclasses can override the method and it would make code navigation easier.

bindings/python/test/test_arrow.py
```diff
@@ -384,3 +388,183 @@ def test_find_decimal128(self):
         )
         table = find_arrow_all(coll, {}, schema=schema)
         self.assertEqual(table, expected)
+
+
+class TestNulls(unittest.TestCase):
```
blink1073, line 393: I'd suggest moving this class to `test/utils.py` and using `raise NotImplementedError` for the methods that need to be overridden.

bindings/python/test/utils.py
```diff
@@ -54,3 +65,186 @@ def succeeded(self, event):
...
+        int64_arr = [(i if (i % 2 == 0) else None) for i in range(len(self.oids))]
+        self.coll.insert_many(
+            [{"_id": self.oids[i], "int64": int64_arr[i]} for i in range(len(self.oids))]
+        )
+
+        table = self.find_fn(self.coll, {}, schema=int_schema)
+
+        # Resulting datatype should be float64 according to the spec for numpy
+        # and pandas.
+        atype = type(self).pytype_tab_map[int]
```
ShaneHarvey, line 165: Here and elsewhere we can remove the `type()` call. You can call access class method/property with `self.pytype_tab_map` instead.

bindings/python/test/test_pandas.py
```diff
@@ -219,3 +224,45 @@ def test_find_decimal128(self):
...
         table = find_pandas_all(self.coll, {}, schema=self.schema)
         pd.testing.assert_frame_equal(expected, table)
+
+
+class TestNulls(TestNullsBase):
+    def find_fn(self, coll, query, schema):
+        return find_pandas_all(coll, query, schema=schema)
+
+    def equal_fn(self, left, right):
+        left = left.fillna(0).replace(-0b1 << 63, 0)  # NaN is sometimes this
```
ShaneHarvey, line 234: What is `-0b1 << 63` and why is it only sometimes?

ShaneHarvey (changes requested):

bindings/python/test/test_pandas.py
bcwarner: On some instances of GitHub Actions, it would fail with the NaNs replaced with a bit string of all 1s instead of 0s and the negative sign (example: https://github.com/mongodb-labs/mongo-arrow/runs/7550724404?check_suite_focus=true).

ShaneHarvey: @blink1073 does this make sense to you?

blink1073: It sounds like an overflow error in pandas, that's a head-scratcher.

ShaneHarvey: That value is actually the minimum int64 value:
```
  File ".../bindings/python/test/test_arrow.py", line 498, in test_int_handling
    self.equal_fn(res_table["int64"], table_write["int64"])
...
AssertionError: Series are different

Series values are different (50.0 %)
[index]: [0, 1, 2, 3]
[left]:  [0, -9223372036854775808, 2, -9223372036854775808]
[right]: [0.0, 0.0, 2.0, 0.0]
```

```
>>> print(bin(-9223372036854775808))
-0b1000000000000000000000000000000000000000000000000000000000000000
```

bindings/python/test/utils.py
```diff
@@ -54,3 +65,186 @@ def succeeded(self, event):
...
+        table_write = self.table_from_dict({"int64": int64_arr})
+
+        write(self.coll, table_write)
+        res_table = self.find_fn(self.coll, {}, schema=int_schema)
+
+        self.equal_fn(res_table["int64"], table_write["int64"])
+        self.assert_in_idx(res_table, "_id")
+        self.assertType(res_table["int64"], atype)
+
+    def test_other_handling(self):
```
ShaneHarvey, line 182: Can we add a test that will automatically fail when we add support for new data types? Something like:
```
def test_all_types():
    for type in _BsonArrowTypes:
        self.assertIn(type, self.pytype_tab_map)
```

bindings/python/test/test_pandas.py
ShaneHarvey: I'm fine if we merge with this hack but could you open a ticket so we can remember to report it as a bug (to pandas or possibly arrow)?

bcwarner: Opened a ticket here: https://jira.mongodb.org/browse/ARROW-113

ShaneHarvey (approved):

PR #84, opened 2022-07-26
ARROW-104 Improve Error Message for Lib ImportError

blink1073:


bindings/python/pymongoarrow/__init__.py
```diff
@@ -29,8 +29,10 @@ def _parse_version(version):
 
 try:
     from pymongoarrow.lib import libbson_version
-except ImportError:
-    warnings.warn("This library has not been compiled")
+except ImportError as e:
+    warnings.warn(
+        f"Could not find compiled pymongoarrow.lib extension, please install from source or report the following traceback on the issue tracker:\n{e}"
+    )
```
ShaneHarvey, line 35: line length

ShaneHarvey: Should we include the full traceback in the warning?

blink1073: Sounds good

bindings/python/pymongoarrow/__init__.py
```diff
@@ -29,10 +30,12 @@ def _parse_version(version):
...
 try:
     from pymongoarrow.lib import libbson_version
-except ImportError as e:
+except ImportError:
     warnings.warn(
-        f"Could not find compiled pymongoarrow.lib extension, please install from source or report the following traceback on the issue tracker:\n{e}"
+        "Could not find compiled pymongoarrow.lib extension, please install "
+        "from source or report the following traceback on the issue tracker"
     )
+    warnings.warn(traceback.print_exc())  # type: ignore[arg-type]
```
ShaneHarvey, line 38: The arg-type is a real bug. This need to use format_exc. Ideally, the format_exc() would be included in the warning above so the user doesn't see two UserWarnings.

blink1073: ACK

ShaneHarvey (approved):
LGTM can you provide an example of what the new behavior looks like?

blink1073:
```python
/Users/steve.silvester/workspace/mongo-arrow/bindings/python/pymongoarrow/__init__.py:34: UserWarning: Could not find compiled pymongoarrow.lib extension, please install from source or report the following traceback on the issue tracker:
  File "/Users/steve.silvester/workspace/mongo-arrow/bindings/python/pymongoarrow/__init__.py", line 32, in <module>
    from pymongoarrow.lib import libbson_version
ModuleNotFoundError: No module named 'pymongoarrow.lib'

  warnings.warn(
```

PR #85, opened 2022-08-05
ARROW-107 Test against Python 3.11 beta

blink1073:


ShaneHarvey (approved):

blink1073:
Looks like we need libarrow to build against since there are no Python 3.11 wheels for PyArrow.

ShaneHarvey:
Ah, let's close this PR and mark the jira ticket blocked?

blink1073:
> Ah, let's close this PR and mark the jira ticket blocked?

That or we could switch to using a conda environment like we do in the docs build and depend on libarrow.

ShaneHarvey:
Sounds like too much effort. We also need to wait for Cython numpy and pandas to support 3.11 too.

blink1073:
Good point, backlogging.

PR #86, opened 2022-08-05
ARROW-106 Support PyArrow 8.0

blink1073:


ShaneHarvey (approved):

ShaneHarvey (changes requested):
Wheels are failing with:
```
ValueError: Cannot repair wheel, because required library "libparquet.so.800" could not be located
```

I suppose parquet is a new dependency  added in Arrow 8 that we need to handle.

blink1073:
> I suppose parquet is a new dependency added in Arrow 8 that we need to handle.

There is a parquet binary in the folder, it just isn't in `pa.get_libraries()`.  I added some special handling.

ShaneHarvey:
> There is a parquet binary in the folder, it just isn't in pa.get_libraries()

Should we report that as a bug?

bindings/python/setup.py
```diff
@@ -158,11 +158,11 @@ def append_arrow_flags(module):
         pass
 
     # Find the appropriate library file and optionally copy it locally.
-    for name in pa.get_libraries():
+    for name in pa.get_libraries() + ["parquet"]:
         if IS_WIN:
             if COPY_LIBARROW:
                 lib_file = os.path.join(arrow_lib, f"{name}.dll")
-                if not os.path.exists(lib_file):
+                if not os.path.exists(lib_file) and name != "parquet":
```
ShaneHarvey, line 165: Won't shutil.copy fail when the parquet lib_file path doesn't exist?

blink1073: Fixed

blink1073:
> Should we report that as a bug?

https://issues.apache.org/jira/browse/ARROW-17327

ShaneHarvey (approved):

PR #88, opened 2022-08-08
ARROW-86 Basic auto-discovery of schemas

bcwarner:
Automatically starts building builders for fields where no schema is present. Currently only tests for arrow, will add the pandas and numpy ones shortly.

### DIFF
--- a/bindings/python/pymongoarrow/api.py
+++ b/bindings/python/pymongoarrow/api.py
@@ -58,7 +58,7 @@ _MAX_MESSAGE_SIZE = 48000000 - 16 * 1024
 _MAX_WRITE_BATCH_SIZE = max(100000, MAX_WRITE_BATCH_SIZE)
 
 
-def find_arrow_all(collection, query, *, schema, **kwargs):
+def find_arrow_all(collection, query, *, schema=None, **kwargs):
     """Method that returns the results of a find query as a
     :class:`pyarrow.Table` instance.
 
@@ -94,7 +94,7 @@ def find_arrow_all(collection, query, *, schema, **kwargs):
     return context.finish()
 
 
-def aggregate_arrow_all(collection, pipeline, *, schema, **kwargs):
+def aggregate_arrow_all(collection, pipeline, *, schema=None, **kwargs):
     """Method that returns the results of an aggregation pipeline as a
     :class:`pyarrow.Table` instance.
 
@@ -145,7 +145,7 @@ def _arrow_to_pandas(arrow_table):
     return arrow_table.to_pandas(split_blocks=True, self_destruct=True)
 
 
-def find_pandas_all(collection, query, *, schema, **kwargs):
+def find_pandas_all(collection, query, *, schema=None, **kwargs):
     """Method that returns the results of a find query as a
     :class:`pandas.DataFrame` instance.
 
@@ -164,7 +164,7 @@ def find_pandas_all(collection, query, *, schema, **kwargs):
     return _arrow_to_pandas(find_arrow_all(collection, query, schema=schema, **kwargs))
 
 
-def aggregate_pandas_all(collection, pipeline, *, schema, **kwargs):
+def aggregate_pandas_all(collection, pipeline, *, schema=None, **kwargs):
     """Method that returns the results of an aggregation pipeline as a
     :class:`pandas.DataFrame` instance.
 
@@ -183,7 +183,7 @@ def aggregate_pandas_all(collection, pipeline, *, schema, **kwargs):
     return _arrow_to_pandas(aggregate_arrow_all(collection, pipeline, schema=schema, **kwargs))
 
 
-def _arrow_to_numpy(arrow_table, schema):
+def _arrow_to_numpy(arrow_table, schema=None):
     """Helper function that converts an Arrow Table to a dictionary
     containing NumPy arrays. The memory buffers backing the given Arrow Table
     may be destroyed after conversion if the resulting Numpy array(s) is not a
@@ -192,16 +192,20 @@ def _arrow_to_numpy(arrow_table, schema):
     See https://arrow.apache.org/docs/python/numpy.html for details.
     """
     container = {}
-    for fname in schema:
-        dtype = get_numpy_type(schema.typemap[fname])
-        if dtype == np.str_:
-            container[fname] = arrow_table[fname].to_pandas().to_numpy(dtype=dtype)
-        else:
-            container[fname] = arrow_table[fname].to_numpy()
+    if schema:
+        for fname in schema:
+            dtype = get_numpy_type(schema.typemap[fname])
+            if dtype == np.str_:
+                container[fname] = arrow_table[fname].to_pandas().to_numpy(dtype=dtype)
+            else:
+                container[fname] = arrow_table[fname].to_numpy()
+    else:
+        for col in arrow_table.column_names:
+            container[col] = arrow_table[col].to_numpy()
     return container
 
 
-def find_numpy_all(collection, query, *, schema, **kwargs):
+def find_numpy_all(collection, query, *, schema=None, **kwargs):
     """Method that returns the results of a find query as a
     :class:`dict` instance whose keys are field names and values are
     :class:`~numpy.ndarray` instances bearing the appropriate dtype.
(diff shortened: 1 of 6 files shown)


bindings/python/test/test_arrow.py
```diff
@@ -325,6 +327,65 @@ def test_string_bool(self):
...
+                }
+            ),
+        )
+
+        # Write this table.if coll is None:
+        coll = self.coll
+        self.coll.drop()
+        res = write(self.coll, data)
+        self.assertEqual(len(data), res.raw_result["insertedCount"])
+        out = find_arrow_all(coll, {}, schema=None).drop(["_id"])
```
ShaneHarvey, line 354: Let's make it so that the schema=None part is optional too.

bindings/python/test/test_arrow.py
```diff
@@ -325,6 +327,65 @@ def test_string_bool(self):
...
+                {
+                    "bool": bool_(),
+                    "dt": timestamp("ms"),
+                    "string": string(),
+                }
+            ),
+        )
+
+        # Write this table.if coll is None:
+        coll = self.coll
```
ShaneHarvey, line 350: is coll needed?

bindings/python/test/test_arrow.py
```diff
@@ -325,6 +327,65 @@ def test_string_bool(self):
...
+        self.coll.drop()
+        res = write(self.coll, data)
+        self.assertEqual(len(data), res.raw_result["insertedCount"])
+        out = find_arrow_all(coll, {}, schema=None).drop(["_id"])
+        self.assertEqual(data, out)
+        # Find it, no schema.
```
ShaneHarvey, line 356: Dangling comment?

bindings/python/test/test_arrow.py
```diff
@@ -325,6 +327,65 @@ def test_string_bool(self):
...
+    def test_auto_schema(self):
+        # Create table with random data of various types.
+        data = Table.from_pydict(
+            {
+                "string": [None] + [str(i) for i in range(2)],
+                "bool": [True for _ in range(3)],
+                "dt": [datetime(1970 + i, 1, 1) for i in range(3)],
+            },
+            # Ordering of output auto-schema will be the first fields detected.
+            # Test will fail if not ordered correctly.
```
ShaneHarvey, line 339: Is this comment only needed because the order of the fields above differs from the schema below?

bindings/python/test/test_arrow.py
```diff
@@ -325,6 +327,65 @@ def test_string_bool(self):
...
+            ArrowSchema(
+                {
+                    "bool": bool_(),
+                    "dt": timestamp("ms"),
+                    "string": string(),
+                }
+            ),
+        )
+
+        # Write this table.if coll is None:
```
ShaneHarvey, line 349: Is this comment still needed?

bindings/python/test/test_arrow.py
```diff
@@ -325,6 +327,65 @@ def test_string_bool(self):
...
+        coll = self.coll
+        self.coll.drop()
+        codec_options = CodecOptions(tzinfo=timezone("US/Eastern"), tz_aware=True)
+        res = write(self.coll.with_options(codec_options=codec_options), data)
+        self.assertEqual(len(data), res.raw_result["insertedCount"])
+        out = find_arrow_all(coll.with_options(codec_options=codec_options), {}, schema=None).drop(
+            ["_id"]
+        )
+        self.assertEqual(data, out)
+        # Find it, no schema.
```
ShaneHarvey, line 387: Dangling comment?

ShaneHarvey (changes requested):
Very cool!

bindings/python/test/test_arrow.py
bcwarner: The columns of the auto-detected schema is in the order of when a field first appears in the stream, so `bool` and `dt` will get detected first, and then `string` third. I had the schema ordered differently before, and it would on `self.assertEqual(data, out)` because the orderings of the schema and the detected schema did not match. I'll also the data dictionary change it to match.

bindings/python/test/test_arrow.py
```diff
@@ -325,6 +327,65 @@ def test_string_bool(self):
             ),
         )
 
+    def test_auto_schema(self):
```
ShaneHarvey, line 330: Let's also add a test with heterogeneous type for the same field, eg:
```python
[{"a": 1},  {"a": "2"},  {"a": True}, {"a": 4}]
```

The test should confirm that we infer from the first type and fill other types with null.

bindings/python/pymongoarrow/api.py
```diff
@@ -190,16 +192,20 @@ def _arrow_to_numpy(arrow_table, schema):
...
+    if schema:
+        for fname in schema:
+            dtype = get_numpy_type(schema.typemap[fname])
+            if dtype == np.str_:
+                container[fname] = arrow_table[fname].to_pandas().to_numpy(dtype=dtype)
+            else:
+                container[fname] = arrow_table[fname].to_numpy()
+    else:
+        for col in arrow_table.column_names:
+            container[col] = arrow_table[col].to_numpy()
```
ShaneHarvey, line 204: Looks like we're loosing the `if dtype == np.str_:` logic. Is that important? Can we add a test case if so?

bindings/python/pymongoarrow/api.py
```diff
@@ -58,7 +58,7 @@
 _MAX_WRITE_BATCH_SIZE = max(100000, MAX_WRITE_BATCH_SIZE)
 
 
-def find_arrow_all(collection, query, *, schema, **kwargs):
+def find_arrow_all(collection, query, *, schema=None, **kwargs):
     """Method that returns the results of a find query as a
```
ShaneHarvey, line 62: For all the changed functions, let's update the docs to say that the `schema` param is optional.

bindings/python/test/test_arrow.py
ShaneHarvey: Ping.

bindings/python/test/test_arrow.py
ShaneHarvey: I still have the same question. If we change the from_pydict and ArrowSchema key order to match, then I believe this comment is no longer needed.

bindings/python/test/test_numpy.py
```diff
@@ -43,14 +43,15 @@ def setUpClass(cls):
...
+    def assert_numpy_equal(self, actual, expected, check_dtype=True):
         self.assertIsInstance(actual, dict)
         for field in expected:
             # workaround np.nan == np.nan evaluating to False
             a = np.nan_to_num(actual[field])
             e = np.nan_to_num(expected[field])
             np.testing.assert_array_equal(a, e)
-            self.assertEqual(actual[field].dtype, expected[field].dtype)
+            if check_dtype:
+                self.assertEqual(actual[field].dtype, expected[field].dtype)
```
ShaneHarvey, line 54: Is check_dtype used anywhere? If not can we remove it?

bindings/python/pymongoarrow/lib.pyx
```diff
@@ -94,6 +103,18 @@ def process_bson_stream(bson_stream, context):
...
+                    # Only run if there is no schema.
+                    ftype = bson_iter_type(&doc_iter)
+                    if ftype not in _builder_type_map:
+                        continue
+                    if _builder_type_map[ftype] == DatetimeBuilder and context.tzinfo is not None:
+                        arrow_type = timestamp(arrow_type.unit, tz=context.tzinfo)
+                        builder_map[key] = _builder_type_map[ftype](dtype=arrow_type)
+                    else:
+                        builder_map[key] = _builder_type_map[ftype]()
+                    builder = builder_map[key]
```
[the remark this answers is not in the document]
ShaneHarvey: Still can remove some `builder_map[key]` lookups here. Also need to remove the `print(context.tzinfo)`.

bindings/python/pymongoarrow/lib.pyx
```diff
@@ -94,6 +103,22 @@ def process_bson_stream(bson_stream, context):
...
+                if builder is None and context.schema is None:
+                    # Only run if there is no schema.
+                    ftype = bson_iter_type(&doc_iter)
+                    if ftype not in _builder_type_map:
+                        continue
+
+                    builder_type = _builder_type_map[ftype]
+                    if builder_type == DatetimeBuilder and context.tzinfo is not None:
+                        arrow_type = timestamp(arrow_type.unit, tz=context.tzinfo)
+                        print(context.tzinfo)
```
ShaneHarvey, line 115: Remove `print`

bindings/python/test/test_numpy.py
```diff
@@ -50,7 +50,6 @@ def assert_numpy_equal(self, actual, expected):
             a = np.nan_to_num(actual[field])
             e = np.nan_to_num(expected[field])
             np.testing.assert_array_equal(a, e)
-            self.assertEqual(actual[field].dtype, expected[field].dtype)
```
ShaneHarvey, line 53: Why remove this assertion?

bcwarner: Oops, forgot that the statement after the `if` I added was there before

ShaneHarvey (approved):
LGTM!

PR #90, opened 2022-08-17
ARROW-115 Changelog for 0.5.0

blink1073:


bindings/python/docs/source/changelog.rst
```diff
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+Changes in Version 0.5.0
+------------------------
+- Support auto-discovery of schemas.
+- Support PyArrow 9.0.
+- Improve Error Message for Lib ImportError.
```
ShaneHarvey, line 8: "Improve error message for lib ..."

bindings/python/docs/source/changelog.rst
```diff
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+Changes in Version 0.5.0
+------------------------
+- Support auto-discovery of schemas.
```
ShaneHarvey, line 6: Worth mentioning the schema param is now option on the various find/aggregate_*_all methods. Also should we mention that it works by inferring the schema from the _first_ document in the result set?

bindings/python/docs/source/changelog.rst
blink1073: Done

bindings/python/docs/source/changelog.rst
blink1073: Done

ShaneHarvey (approved):

PR #92, opened 2022-08-25
ARROW-117 "process_bson_stream" Does Not Construct Projection With Auto-Discovered Schema

juliusgeo:


blink1073:
I think we should replicate the auto schema tests in `test_numpy.py` and `test_pandas.py` as well to prevent regression.

blink1073 (approved):
LGTM!

ShaneHarvey (approved):

PR #95, opened 2022-08-30
ARROW-112 Make PyArrow Binding more Robust

blink1073:


ShaneHarvey:
Can you explain why this is more robust?

blink1073:
We are using the linker to get the files, which automatically brings in any dependent arrow libraries, as opposed to finding the files by name ourselves.

bindings/python/repair_wheel.py
```diff
@@ -0,0 +1,80 @@
...
+    print(f"{name}: {os.environ[name]}")
+
+
+if os.name == "nt":
+    append_os_variable("PATH")
+    run([sys.executable, "-m", "pip", "install", "delvewheel"])
+    run(["delvewheel", "repair", "--no-mangle", "ucrtbased.dll", "-w", wheel_dir, wheel_file])
+
+elif sys.platform == "darwin":
+    # FIXME: We should not have to do this.
```
ShaneHarvey, line 61: Are you hoping to fix this in this PR?

bindings/python/repair_wheel.py
```diff
@@ -0,0 +1,80 @@
...
+HERE = os.path.abspath(os.path.dirname(__file__))
+wheel_dir, wheel_file, delocate_args = sys.argv[1:]
+wheel_dir = wheel_dir.replace(os.sep, "/")
+wheel_file = wheel_file.replace(os.sep, "/")
+
+# Ensure pyarrow.
+if "universal2" in wheel_file:
+    macos_ver = os.environ.get("MACOSX_DEPLOYMENT_TARGET", "10.3")
+    macos_ver = macos_ver.replace(".", "_")
+    target = os.path.expanduser("~/wheels")
```
ShaneHarvey, line 15: Using the home directory (`~`) seems off.

bindings/python/repair_wheel.py
```diff
@@ -0,0 +1,80 @@
...
+    libbson_lib = glob.glob(os.path.join(libbson, "bin"))
+else:
+    libbson_lib = glob.glob(os.path.join(libbson, "lib*"))
+extra_paths = pa.get_library_dirs() + libbson_lib
+extra_path = os.path.pathsep.join([a.replace(os.sep, "/") for a in extra_paths])
+
+
+def append_os_variable(name):
+    if os.environ.get(name):
+        os.environ[name] = os.environ[name] + os.pathsep + extra_path
```
ShaneHarvey, line 49: Can you make extra_path an argument? Mixing a global in this method makes it harder to read imo.

bindings/python/repair_wheel.py
blink1073: I'm waiting to hear if they want me to fix it in Arrow.

bindings/python/repair_wheel.py
blink1073: I'll use a tempdir here

bindings/python/repair_wheel.py
blink1073: Done

bindings/python/repair_wheel.py
```diff
@@ -0,0 +1,83 @@
...
+import tempfile
+from subprocess import run
+
+HERE = os.path.abspath(os.path.dirname(__file__))
+wheel_dir, wheel_file, delocate_args = sys.argv[1:]
+wheel_dir = wheel_dir.replace(os.sep, "/")
+wheel_file = wheel_file.replace(os.sep, "/")
+
+# Ensure pyarrow.
+if "universal2" in wheel_file:
```
ShaneHarvey, line 14: Why can't we rely on regular pip install like we do for other platforms? Could you add a comment?

bindings/python/repair_wheel.py
```diff
@@ -0,0 +1,83 @@
...
+        os.environ[name] = os.environ[name] + os.pathsep + extra_path
+    else:
+        os.environ[name] = extra_path
+    print(f"{name}: {os.environ[name]}")
+
+
+if os.name == "nt":
+    append_os_variable("PATH", extra_path)
+    run([sys.executable, "-m", "pip", "install", "delvewheel"])
+    run(["delvewheel", "repair", "--no-mangle", "ucrtbased.dll", "-w", wheel_dir, wheel_file])
```
ShaneHarvey, line 61: What is ucrtbased.dll? Why does it need to be hardcoded?

ShaneHarvey (commented):
I worry how robust this actually is given we're adding dependencies on auditwheel, delocate, and delvewheel.

blink1073:
> I worry how robust this actually is given we're adding dependencies on auditwheel, delocate, and delvewheel.

Auditwheel and delocate are used by default in cibuildwheel, and they recommend using delvewheel [if needed](https://cibuildwheel.readthedocs.io/en/stable/options/#repair-wheel-command).

bindings/python/repair_wheel.py
blink1073: Done

bindings/python/repair_wheel.py
blink1073: Done

ShaneHarvey (approved):

PR #99, opened 2022-11-10
ARROW-120 Support PyArrow 10.0

blink1073:


bindings/python/setup.py
```diff
@@ -147,15 +147,17 @@ def append_arrow_flags(ext):
...
-        ext.extra_compile_args.append("-std=c++11")
+        ext.extra_compile_args.append("-std=c++17")
 
     # Arrow's manylinux{2010, 2014} binaries are built with gcc < 4.8 which predates CXX11 ABI
     # - https://uwekorn.com/2019/09/15/how-we-build-apache-arrows-manylinux-wheels.html
     # - https://arrow.apache.org/docs/python/extending.html#example
-    if "std=" not in os.environ.get("CXXFLAGS", ""):
-        ext.extra_compile_args.append("-std=c++11")
+    if os.name != "nt" and "std=" not in os.environ.get("CXXFLAGS", ""):
         ext.extra_compile_args.append("-D_GLIBCXX_USE_CXX11_ABI=0")
```
ShaneHarvey, line 156: Was the old GLIBCXX_USE_CXX11_ABI behavior broken on windows?

blink1073: No, but it is only used on Linux, per the blog post.  Windows requires a different argument for setting the c++ standard, so I refactored

bindings/python/setup.py
```diff
@@ -147,15 +150,17 @@ def append_arrow_flags(ext):
...
 
+    if os.name == "nt":
+        ext.extra_compile_args.append("/std:c++17")
```
ShaneHarvey, line 162: Could we move this up so that it's next to the `if os.name == "posix":`?

blink1073: I grouped things a bit better now

ShaneHarvey (approved):

PR #105, opened 2022-11-16
ARROW-126 pymongoarrow no longer works when pandas is not installed

blink1073:


bindings/python/pymongoarrow/api.py
```diff
@@ -17,10 +17,19 @@
...
 from bson.raw_bson import RawBSONDocument
-from numpy import ndarray
-from pandas import DataFrame
 from pyarrow import Schema as ArrowSchema
 from pyarrow import Table
+
+try:
+    from numpy import ndarray
+except ImportError:
+    ndarray = None
```
ShaneHarvey, line 26: Numpy is always installed so this is not needed:
```
$ pip install ~/Downloads/all-dist/pymongoarrow-0.7.0.dev0-cp37-cp37m-macosx_10_14_x86_64.whl
Processing ./Downloads/all-dist/pymongoarrow-0.7.0.dev0-cp37-cp37m-macosx_10_14_x86_64.whl
Collecting pyarrow<10.1,>=10
  Using cached pyarrow-10.0.0-cp37-cp37m-macosx_10_14_x86_64.whl (24.6 MB)
Collecting pymongo<5,>=3.11
  Using cached pymongo-4.3.2-cp37-cp37m-macosx_10_6_intel.whl (410 kB)
Collecting numpy>=1.16.6
  Using cached numpy-1.21.6-cp37-cp37m-macosx_10_9_x86_64.whl (16.9 MB)
Collecting dnspython<3.0.0,>=1.16.0
  Using cached dnspython-2.2.1-py3-none-any.whl (269 kB)
Installing collected packages: numpy, pyarrow, dnspython, pymongo, pymongoarrow
Successfully installed dnspython-2.2.1 numpy-1.21.6 pyarrow-10.0.0 pymongo-4.3.2 pymongoarrow-0.7.0.dev0
```

ShaneHarvey (changes requested):

bindings/python/pymongoarrow/api.py
blink1073: I don't think we should rely on a transient dependency being available.

ShaneHarvey (approved):

PR #106, opened 2022-11-17
ARROW-127 Time Zones are Not Properly Handled by Auto Schema

blink1073:
Also handles a few warnings and adds warning handling.

### DIFF
--- a/bindings/python/pymongoarrow/types.py
+++ b/bindings/python/pymongoarrow/types.py
@@ -75,7 +75,10 @@ _TYPE_NORMALIZER_FACTORY = {
     Int64: lambda _: int64(),
     float: lambda _: float64(),
     int: lambda _: int64(),
-    datetime: lambda _: timestamp("ms"),  # TODO: add tzinfo support
+    # Note: we cannot infer a timezone form a raw datetime class,
+    # if a timezone is preferred then a timestamp with tz information
+    # must be used directly.
+    datetime: lambda _: timestamp("ms"),
     ObjectId: lambda _: ObjectIdType(),
     Decimal128: lambda _: Decimal128StringType(),
     str: lambda _: string(),
--- a/bindings/python/pymongoarrow/types.py
+++ b/bindings/python/pymongoarrow/types.py
@@ -88,7 +88,7 @@ _TYPE_CHECKER_TO_NUMPY = {
     _atypes.is_int64: np.int64,
     _atypes.is_float64: np.float64,
     _atypes.is_timestamp: "datetime64[ms]",
-    _is_objectid: np.object,
+    _is_objectid: object,
     _atypes.is_string: np.str_,
     _atypes.is_boolean: np.bool_,
 }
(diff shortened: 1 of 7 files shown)


bindings/python/pymongoarrow/types.py
```diff
@@ -88,7 +88,7 @@ def _is_decimal128_str(obj):
     _atypes.is_int64: np.int64,
     _atypes.is_float64: np.float64,
     _atypes.is_timestamp: "datetime64[ms]",
-    _is_objectid: np.object,
+    _is_objectid: object,
```
ShaneHarvey, line 91: What is the deal with the comment on line 78?:
```
    datetime: lambda _: timestamp("ms"),  # TODO: add tzinfo support
```

blink1073: That was pre-existing, I plan to look at it as part of this PR.

blink1073: It turns out there was nothing to do, I added a note.

bindings/python/pymongoarrow/types.py
```diff
@@ -75,7 +75,10 @@ def _is_decimal128_str(obj):
     Int64: lambda _: int64(),
     float: lambda _: float64(),
     int: lambda _: int64(),
-    datetime: lambda _: timestamp("ms"),  # TODO: add tzinfo support
+    # Note: we cannot infer a timezone form a raw datetime class,
+    # if a timezone is preferred then a timestamp with tz information
+    # must be used directly.
+    datetime: lambda _: timestamp("ms"),
```
ShaneHarvey, line 81: What is this used for? If we have a collection with tz_aware=True, that collection will produce UTC timezone aware datetimes. If we call find_arrow_all with the same collection, I would expect the arrow datetimes to also be UTC timezone aware. Is that not the case? Perhaps the schema discovery needs to take the collection's CodecOptions into account?

blink1073: The schema discovery does take into account the codec options.   This particular code path is only hit when the user explicitly uses the `datetime` class as the type in a schema.

ShaneHarvey: Ooooooh okay. I guess that makes sense to me. Seems like a bit of a pitfall but fair enough.

ShaneHarvey (approved):

bindings/python/pymongoarrow/types.py
blink1073: We could return a special type here that signifies "tz unknown" and then finalize it when we create the context.

ShaneHarvey: I think the current behavior is okay for now. Could you create a ticket for that idea though? I have some more thoughts but don't want to hold up this PR.

blink1073: https://jira.mongodb.org/browse/ARROW-132

PR #108, opened 2022-11-18
ARROW-128 Allow tests to be run with PyTest

blink1073:


.github/workflows/test-python.yml
```diff
@@ -65,22 +69,26 @@ jobs:
...
         run: |
           # Install the test deps
-          LIBBSON_INSTALL_DIR=$(pwd)/libbson python -m pip install -vv -e ".[test]"
-          LD_LIBRARY_PATH=$(pwd)/libbson/lib python -m unittest discover test -v
+          python -m pip install -vv -e ".[test]"
+          python -m unittest discover test -v
+      - name: Run the tests with pytest
+        run: |
+          pip install pytest
+          pytest -vv
```
ShaneHarvey, line 91: I don't think it makes sense to run the test suite twice. We should only support and only run one test framework.

bindings/python/test/test_arrow.py
```diff
@@ -16,7 +16,7 @@
 import unittest.mock as mock
 from datetime import datetime
 from test import client_context
-from test.utils import AllowListEventListener, TestNullsBase
+from test.utils import AllowListEventListener, TullsTestMixin
```
ShaneHarvey, line 19: TullsTestMixin -> NullsTestMixin?

ShaneHarvey (changes requested):

.github/workflows/test-python.yml
blink1073: It only takes 21 seconds, and I didn't want to impose `pytest`, but it is largely compatible with `unittest`.  This guard was meant to ensure that either would continue to work.

bindings/python/test/test_arrow.py
blink1073: Fixed

.github/workflows/test-python.yml
ShaneHarvey: It still is not worthwhile to maintain compatibility with two different test runners. If we want to switch to pytest then we should drop unittest and only use pytest (not that I'm suggesting it).

PR #118, opened 2022-12-20
ARROW-134 Cannot encode pandas NA objects

blink1073:


bindings/python/pymongoarrow/api.py
```diff
@@ -316,6 +318,21 @@ def _tabular_generator(tabular):
...
+    python_type = NA.__class__  # type:ignore[assignment]
+    bson_type = None  # type:ignore[assignment]
+
+    def transform_python(self, _):
+        """Transform an NA object into 'None'"""
+        return None
+
+    def transform_bson(self, _):
+        """Transform a 'None' object into NA"""
+        return NA
```
ShaneHarvey, line 333: IIUC this is only used for write() so the decoding side is never used. Can we remove transform_bson and only use the TypeEncoder class for clarity?

bindings/python/pymongoarrow/api.py
```diff
@@ -316,6 +318,21 @@ def _tabular_generator(tabular):
             return
 
 
+class _PandasNACodec(TypeCodec):
+    """A custom type codec for Pandas NA objects."""
+
+    python_type = NA.__class__  # type:ignore[assignment]
+    bson_type = None  # type:ignore[assignment]
```
ShaneHarvey, line 325: Interesting, why the type ignores?

ShaneHarvey (changes requested):

bindings/python/pymongoarrow/api.py
```diff
@@ -352,6 +369,13 @@ def write(collection, tabular):
         )
 
     tabular_gen = _tabular_generator(tabular)
+
+    # Handle Pandas NA objects.
+    codec_options = collection.codec_options
+    if DataFrame is not None:
+        type_registry = TypeRegistry([_PandasNACodec()])
+        codec_options = codec_options.with_options(type_registry=type_registry)
```
ShaneHarvey, line 377: I'm not sure it's ideal to completely replace the collections type_registry. What if the app already configured a type_registry for other types they want to encode?

Would it be possible to keep the existing registry but add the _PandasNACodec?

bindings/python/pymongoarrow/api.py
blink1073: Done

bindings/python/pymongoarrow/api.py
blink1073: My editor had flagged them because the base class uses read-only properties.

bindings/python/pymongoarrow/api.py
blink1073: We could, but we'd have to use private APIs from `TypeRegistry`.

bindings/python/pymongoarrow/api.py
ShaneHarvey: I see mypy does not raise this issue. This does look like a bug in pyright though:
```
  /Users/shane/git/mongo-python-driver/ignore.py:15:19 - error: Expression of type "python_type" cannot be assigned to declared type "property"
    "Type[type]" is incompatible with "Type[property]" (reportGeneralTypeIssues)
```

Looks like https://github.com/microsoft/pyright/issues/2601 which they closed as won't fix.

ShaneHarvey: Actually that pyright issue has `@classmethod` whereas we just use `@property` so I think it would be worth opening a new pyright issue for it.

bindings/python/test/test_pandas.py
```diff
@@ -98,13 +99,21 @@ def test_aggregate_simple(self):
         self.assertEqual(agg_cmd.command["pipeline"][0]["$project"], projection)
         self.assertEqual(agg_cmd.command["pipeline"][1]["$project"], {"_id": True, "data": True})
 
+    def _assert_frames_equal(self, incoming, outgoing):
+        for name in incoming.columns:
+            col = incoming[name]
+            val = outgoing[name]
+            if str(val.dtype) in ["object", "float64"]:
+                val = val.astype(col.dtype)
+            pd.testing.assert_series_equal(col, val)
```
ShaneHarvey, line 108: Is using "val" as the name of a column idomatic?

bindings/python/test/test_pandas.py
```diff
@@ -98,13 +99,21 @@ def test_aggregate_simple(self):
         self.assertEqual(agg_cmd.command["pipeline"][0]["$project"], projection)
         self.assertEqual(agg_cmd.command["pipeline"][1]["$project"], {"_id": True, "data": True})
 
+    def _assert_frames_equal(self, incoming, outgoing):
+        for name in incoming.columns:
+            col = incoming[name]
+            val = outgoing[name]
+            if str(val.dtype) in ["object", "float64"]:
```
ShaneHarvey, line 106: A comment here or a method docstring would be helpful.

bindings/python/test/test_pandas.py
```diff
@@ -129,23 +138,34 @@ def _create_data(self):
...
     def test_write_schema_validation(self):
         arrow_schema, data = self._create_data()
+
+        # Work around https://github.com/pandas-dev/pandas/issues/11453.
+        def new_replace(k):
+            if k.value < 1:
+                return datetime.datetime(1970, 1, 1)
+            return k.replace(tzinfo=None)
+
+        data["datetime"] = data.apply(lambda row: new_replace(row["datetime"]), axis=1)
```
ShaneHarvey, line 167: That pandas ticket seems to be about .time() not working on NA types but we don't use .time() anywhere. Can you explain?

ShaneHarvey (changes requested):

bindings/python/pymongoarrow/api.py
blink1073: https://jira.mongodb.org/browse/ARROW-135

bindings/python/pymongoarrow/api.py
blink1073: https://github.com/microsoft/pyright/issues/4364

bindings/python/test/test_pandas.py
blink1073: Fixed

bindings/python/test/test_pandas.py
blink1073: Done

bindings/python/test/test_pandas.py
blink1073: Ah, meant to link to https://github.com/pandas-dev/pandas/issues/16248.  Also added context.

ShaneHarvey: I'm still confused since this code is changing the data _before_ passing it to an pymongoarrow method. What happens if the user actually calls write() with a NA datetime? And why are we clamping the datetime to `datetime.datetime(1970, 1, 1)`?

blink1073: They would see the same error as https://jira.mongodb.org/browse/FREE-165786, for which this is a workaround.  I just chose a random valid date.

ShaneHarvey: Thanks, 2 follow up questions.
1) Can we change this to something like?:
```python
if isinstance(k, Nat):
   return None
return k
```

2) this looks like something we should ideally be working around in write() itself. Can you open a ticket for making write() encode NaT datetimes as BSON null?

blink1073: If we return `None` we get the original error, so I had to return the workaround datetime object.  I filed https://jira.mongodb.org/browse/ARROW-136.

ShaneHarvey (approved):

PR #120, opened 2023-01-05
ARROW-55 Allow specifying a custom type for converting ExtensionScalar to python object

juliusgeo:


bindings/python/test/test_schema.py
```diff
@@ -14,12 +14,36 @@
...
+from bson import Decimal128, Int64, ObjectId
+from pyarrow import Table, float64, int64
+from pyarrow import schema as ArrowSchema
+from pyarrow import timestamp
 from pymongoarrow.schema import Schema
+from pymongoarrow.types import _TYPE_NORMALIZER_FACTORY
 
 
 class TestSchema(TestCase):
+    def test_as_py(self):
```
blink1073, line 26: I'm having trouble following along with this test.  I think some comments and more descriptive variable names are needed.  Also, it looks like you're using `_id` for all of the dtypes, perhaps that should be called `value`?

blink1073 (approved):
Thanks!

PR #121, opened 2023-01-05
ARROW-137 Remove workarounds for extension types in test suite

juliusgeo:


bindings/python/test/test_arrow.py
```diff
@@ -244,27 +244,43 @@ def test_pymongo_error(self):
             exc.exception.details.keys(), {"nInserted", "writeConcernErrors", "writeErrors"}
         )
 
-    def _create_data(self):
+    def _create_data(self, block=False, nested=True):
```
blink1073, line 247: What does "block" mean?  It looks like this is only ever called with `block=True`.  When would we not want to include the extension types?

blink1073 (changes requested):
There is also a `.DS_Store` file committed.

bindings/python/test/utils.py
```diff
@@ -145,7 +145,10 @@ def setUp(self):
 
     def assertType(self, obj1, arrow_type):
         if isinstance(obj1, pyarrow.ChunkedArray):
-            if "storage_type" in dir(arrow_type):
+            if "storage_type" in dir(arrow_type) and arrow_type not in (
+                ObjectIdType(),
```
blink1073, line 149: Perhaps we can check for an extension type here so we don't have to manually update?

ShaneHarvey:
Should we close this PR now based on the extra work required?

PR #123, opened 2023-01-12
ARROW-142 Update documentation to make clear that extension types are not fully supported

juliusgeo:


bindings/python/docs/source/extension_types.rst
```diff
@@ -0,0 +1,40 @@
...
+Extension types with Arrow
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+Both extension types, ObjectId and Decimal128, are supported in Arrow. This means that when using
+a PyArrow table, you can use the corresponding PyMongoArrow extension type like so::
+
+        schema = Schema({"_id": ObjectIdType(), "data": Decimal128StringType()})
+        table = find_arrow_all(coll, {}, schema=schema)
+        print(table)
+        >>> pyarrow.Table
+        >>> _id: fixed_size_binary[12]
```
blink1073, line 18: These still aren't extension types, since we're not currently casting the response to `ObjectIdType()`.

juliusgeo: Yup, we can't make them true arrow extension types without implementing pandas extension types. All of that work has been deferred.

blink1073: Can you please update the title of the PR and the ticket to say "not fully supported"?

blink1073:
There's a docs error:

> /home/runner/work/mongo-arrow/mongo-arrow/bindings/python/docs/source/extension_types.rst:document isn't included in any toctree

bindings/python/docs/source/extension_types.rst
juliusgeo: Done.

bindings/python/docs/source/supported_types.rst
```diff
@@ -6,6 +6,10 @@ Supported Types
 PyMongoArrow currently supports a small subset of all BSON types.
 Support for additional types will be added in subsequent releases.
 
+.. note:: PyMongoArrow does not currently support extension types with Pandas. This means that ObjectId
```
blink1073, line 9: This section needs to be updated as well

blink1073 (changes requested):

bindings/python/docs/source/supported_types.rst
```diff
@@ -6,6 +6,11 @@ Supported Types
 PyMongoArrow currently supports a small subset of all BSON types.
 Support for additional types will be added in subsequent releases.
 
+.. note:: PyMongoArrow does not currently support extension types with Pandas/NumPy or Arrow. However, they can be used in schemas.
+   This means that ObjectId and Decimal128 are not supported in Pandas DataFrames or Arrow Tables.
+   Instead, the schema type will be converted to a string representation of the type.
```
blink1073, line 11: It is binary for object id, and str for decimal128

juliusgeo: I was considering it from the perspective of there being {fixed_width_binary(12), string, object, object}, so in this intro I would just describe the general case.

bindings/python/docs/source/extension_types.rst
```diff
@@ -0,0 +1,40 @@
+Extension Types
+===============
+
+This tutorial is intended as an introduction to working with
```
blink1073, line 4: We should include a link to the PyArrow extension type docs

bindings/python/docs/source/extension_types.rst
```diff
@@ -0,0 +1,40 @@
...
+        >>> _id: [[63C003BF0A1D5281D33B0AFD,63C003BF0A1D5281D33B0AFE,63C003BF0A1D5281D33B0AFF,63C003BF0A1D5281D33B0B00]]
+        >>> data: [["0.1","1.0","0.00001",null]]
+        >>> ...
+
+
+
+Extension types with Pandas/NumPy
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Extension types with Pandas/NumPy are only partially supported. They will work when used in a
+schema, but will show up in the table as a :class:`pandas.object`::
```
blink1073, line 30: I think we should also say that the underlying value is ``fixed_size_binary(12)` or `string` as above.

blink1073 (changes requested):

bindings/python/docs/source/extension_types.rst
juliusgeo: In Pandas/NumPy that is not the case. The underlying type is object, because we don't have pandas extension types like with Arrow.

blink1073: The array type is object, but when you access a field it will be a python object.

juliusgeo: Perhaps I'm misunderstanding, but isn't array type == "will show up in the table as"?

juliusgeo: That makes sense. Added.

bindings/python/docs/source/supported_types.rst
```diff
@@ -6,6 +6,11 @@ Supported Types
 PyMongoArrow currently supports a small subset of all BSON types.
 Support for additional types will be added in subsequent releases.
 
+.. note:: PyMongoArrow does not currently support extension types with Pandas/NumPy or Arrow. However, they can be used in schemas.
+   This means that ObjectId and Decimal128 are not supported in Pandas DataFrames or Arrow Tables.
```
blink1073, line 10: "not fully supported" here as well?

juliusgeo: ObjectId and Decimal128 are not supported in Pandas, only "bytes" and "str". This distinction is important because it will dictate how operations are applied to that column.

blink1073: What I mean is that you can still use those types in your schemas and documents, they just have limitations.

juliusgeo: Ah, I see what you mean. Changed.

blink1073 (approved):
Thanks!

ShaneHarvey (approved):

PR #126, opened 2023-02-01
ARROW-143 Switch to Pytest as Test Runner

blink1073:


.github/workflows/test-python.yml
```diff
@@ -65,22 +69,22 @@ jobs:
...
+          python -m pip install -vv -e "."
       - name: Ensure imports with no test deps
         run: |
           python -c "from pymongoarrow.monkey import patch_all; patch_all()"
       - name: Run tests
         run: |
           # Install the test deps
-          LIBBSON_INSTALL_DIR=$(pwd)/libbson python -m pip install -vv -e ".[test]"
-          PYTHONWARNINGS=error LD_LIBRARY_PATH=$(pwd)/libbson/lib python -m unittest discover test -v
+          python -m pip install -vv -e ".[test]"
```
ShaneHarvey, line 86: The `-vv` here and above causes thousands of lines of pip output which makes it difficult to look at the test runs. Can we remove it completely (or use `-v` if that lowers the verbosity)? Or maybe we can use `-v` for the "Install pymongoarrow" part and no `-v` for the "Run tests" part?

ShaneHarvey (changes requested):

.github/workflows/test-python.yml
blink1073: Updated

ShaneHarvey (approved):

PR #128, opened 2023-02-06
ARROW-52 Add support for BSON binary type

blink1073:


bindings/python/pymongoarrow/lib.pyx
```diff
@@ -93,6 +98,10 @@ cdef extract_field_dtype(bson_iter_t * doc_iter, bson_iter_t * child_iter, bson_
         field_type = extract_document_dtype(child_iter, context)
     elif value_t == BSON_TYPE_DATE_TIME:
         field_type = timestamp('ms', tz=context.tzinfo)
+    elif value_t == BSON_TYPE_BINARY:
+        bson_iter_binary (doc_iter, &subtype,
+                            &val_buf_len, &val_buf)
```
ShaneHarvey, line 103: nit: this could fit on one line

bindings/python/pymongoarrow/lib.pyx
```diff
@@ -257,20 +270,25 @@ def process_bson_stream(bson_stream, context, arr_value_builder=None):
...
                             raise ValueError("Subarray is invalid")
-                        builder.append(<bytes>doc_buf[:doc_buf_len])
+                        builder.append(<bytes>val_buf[:val_buf_len])
                     else:
                         builder.append_null()
+                elif ftype == t_binary:
+                    if value_t == BSON_TYPE_BINARY:
+                        bson_iter_binary (&doc_iter, &subtype,
+                            &val_buf_len, &val_buf)
+                        builder.append(<bytes>val_buf[:val_buf_len])
```
ShaneHarvey, line 291: Shouldn't we validate that the subtype is the expected one? Could you add a test for reading data with mixed subtypes?

ShaneHarvey (changes requested):

bindings/python/pymongoarrow/lib.pyx
blink1073: Fixed

bindings/python/pymongoarrow/lib.pyx
blink1073: Done and done

bindings/python/pymongoarrow/lib.pyx
```diff
@@ -288,6 +287,8 @@ def process_bson_stream(bson_stream, context, arr_value_builder=None):
                     if value_t == BSON_TYPE_BINARY:
                         bson_iter_binary (&doc_iter, &subtype,
                             &val_buf_len, &val_buf)
+                        if subtype != builder.subtype:
+                            raise ValueError(f'Expected Binary subtype {builder.subtype}, got {subtype}')
```
ShaneHarvey, line 291: Interesting. Based on our current behavior I would have expected this to be an append_null() in the same way that a non-Binary value would lead to an append_null().

blink1073: Actually that brings up a good point about complex types.  We haven't added a test where you have an embedded document with missing or extra keys, what should be the behavior there?  By this same logic it seems like we should ignore extra keys and null out empty keys.

blink1073: Also if a key's value is of the wrong type, I presume that should be null.

ShaneHarvey: Right, I believe you are describing the existing behavior. Can you confirm?

blink1073: I added a confirming test for the embedded document behavior and updated this behavior to match.

ShaneHarvey (approved):

PR #129, opened 2023-02-09
ARROW-139 Fully support ObjectId extension type

blink1073:


bindings/python/test/test_bson.py
```diff
@@ -120,7 +120,7 @@ def test_simple(self):
             {"_id": ids[2], "dätá": 30},
             {"_id": ids[3], "dätá": 40},
         ]
-        as_dict = {"_id": [oid.binary for oid in ids], "dätá": [10, 20, 30, 40]}
+        as_dict = {"_id": [oid for oid in ids], "dätá": [10, 20, 30, 40]}
```
ShaneHarvey, line 123: `[oid for oid in ids]` -> `ids`

bindings/python/test/test_pandas.py
```diff
@@ -344,9 +344,9 @@ def setUp(self):
 
     def test_find_decimal128(self):
         decimals = [str(i) for i in self.decimal_128s] + [None]  # type:ignore
-        pd_schema = {"_id": np.object_, "decimal128": np.object_}
+        pd_schema = {"_id": PandasBSONObjectId(), "decimal128": np.object_}
         expected = pd.DataFrame(
-            data={"_id": [i.binary for i in self.oids], "decimal128": decimals}
+            data={"_id": [i for i in self.oids], "decimal128": decimals}
```
ShaneHarvey, line 349: `[i for i in self.oids]` -> `self.oids`

ShaneHarvey (changes requested):

ShaneHarvey:
Do we need to updated the docs in this PR (https://mongo-arrow.readthedocs.io/en/latest/extension_types.html) or is there another ticket for that?

blink1073:
I opened https://jira.mongodb.org/browse/ARROW-151

bindings/python/test/test_bson.py
blink1073: Done

bindings/python/test/test_pandas.py
blink1073: Done

juliusgeo:
I think it would be good if you also removed the workarounds for `ObjectId` types in the test suite: 
```
schema = {
            k.__name__: v(True)
            for k, v in _TYPE_NORMALIZER_FACTORY.items()
            if k.__name__ not in ("ObjectId", "Decimal128", "Binary")
        }
```
We have a ticket to remove these workarounds, but I think it would be easier for future us if we remove the workarounds in the same PR as adding support. Then, it will be easier to see which changes were related to specifically removing the ObjectId workaround, rather than just one commit for all of the workarounds.

ShaneHarvey (approved):
LGTM after Julius' comment is addressed.

blink1073:
Addressed

bindings/python/test/test_pandas.py
```diff
@@ -376,7 +375,7 @@ def assert_in_idx(self, table, col_name):
         int: ["int64", "float64"],
         float: "float64",
         datetime.datetime: "datetime64[ns]",
-        ObjectId: "object",
+        ObjectId: "bson_PandasBSONObjectId",
```
juliusgeo, line 378: What do you think about renaming these types to just `PandasObjectId` or similar so that the names become `bson_PandasObjectId` rather than repeating BSON twice?

blink1073: Done

juliusgeo (approved):
LGTM! What a relief to start removing the workarounds :)

PR #135, opened 2023-02-28
Add documentation on how to deal with nulls in data and how to flatten data.

sibbiii:
Hi,

It took me some time to figure out how to flatten data extracted by pymongoarrow efficiently.
Also its not well documented how to deal with null-able datatypes.
So I decided to give back to the community by adding guidance in the documentation.

Maybe you can also add this functionality to the aggregate_pandas_all() convenience function some time ...

Greetings from Austria,
Sebastian

### DIFF
--- a/bindings/python/docs/source/schemas.rst
+++ b/bindings/python/docs/source/schemas.rst
@@ -51,7 +51,7 @@ For Pandas and NumPy you can do the same exact thing:
    0  string   {'start': 0}
    1  string  {'start': 10}
 
-To get a flat pandas dataframe with, e.g., "prop.name" and "prop.start", you can flatten the arrow_table before conversion
+To get a flat Pandas DataFrame with, e.g., "prop.name" and "prop.start", you can flatten the arrow_table before conversion
 
 .. code-block:: pycon
 
--- a/bindings/python/docs/source/schemas.rst
+++ b/bindings/python/docs/source/schemas.rst
@@ -51,6 +51,12 @@ For Pandas and NumPy you can do the same exact thing:
    0  string   {'start': 0}
    1  string  {'start': 10}
 
+To get a flat pandas dataframe with, e.g., "prop.name" and "prop.start", you can flatten the arrow_table before conversion
+
+.. code-block:: pycon
+
+   >>> df = arrow_table.flatten().to_pandas(split_blocks=True, self_destruct=True)
+   ... del arrow_table
 
 Nested Data With Projections
 ----------------------------
--- a/bindings/python/docs/source/supported_types.rst
+++ b/bindings/python/docs/source/supported_types.rst
@@ -53,3 +53,33 @@ respectively, and '_id' that is an `ObjectId`, your schema can be defined as::
 
 Unsupported data types in a schema cause a ``ValueError`` identifying the
 field and its data type.
+
+Null Values and Conversion to Pandas DataFrames
+-----------------------------------------------
+
+In Arrow, all Arrays are always nullable. 
+Pandas has experimental nullable data types as, e.g., "Int64" (note the capital "I").
+You can instruct Arrow to create a pandas DataFrame using nullable dtypes 
+with the code below (taken from `here <https://arrow.apache.org/docs/python/pandas.html>`_)
+
+.. code-block:: pycon
+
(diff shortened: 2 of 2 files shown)


juliusgeo:
Hi! Thank you for this PR. We currently do mention the best way to flatten data when using PyMongoArrow, but we may have hidden it by putting it in the `Schema` page: https://mongo-arrow.readthedocs.io/en/latest/schemas.html. (TLDR you can use MongoDB projections to flatten it on the server).
As for the nullable data-types, could you edit it to a also mention the two existing issues we have for Null types: https://jira.mongodb.org/browse/ARROW-113 and https://jira.mongodb.org/browse/ARROW-136?

bindings/python/docs/source/schemas.rst
```diff
@@ -51,6 +51,12 @@ For Pandas and NumPy you can do the same exact thing:
    0  string   {'start': 0}
    1  string  {'start': 10}
 
+To get a flat Pandas DataFrame with, e.g., "prop.name" and "prop.start", you can flatten the arrow_table before conversion
+
+.. code-block:: pycon
+
+   >>> df = arrow_table.flatten().to_pandas(split_blocks=True, self_destruct=True)
+   ... del arrow_table
```
ShaneHarvey, line 59: Neat! This sounds like it deserves it's own section header, like:
```
Flatten Nested Data with pyarrow.Table.flatten()
------------------------------------------------

One can flatten an arrow table using :class:`pyarrow.Table.flatten`, for example with "prop.name" and "prop.start":

.. code-block:: pycon

   >>> arrow_table = find_arrow_all(coll, {})
   >>> arrow_table = arrow_table.flatten()
   pyarrow.Table
   start: string
   prop.name: string
   prop.start: int32
   ----
   start: [["string","string"]]
   prop.name: [["foo","bar"]]
   prop.start: [[0,10]]


To get a flat Pandas DataFrame, call :class:`pyarrow.Table.flatten` before conversion:

.. code-block:: pycon

   >>> df = arrow_table.flatten().to_pandas(split_blocks=True, self_destruct=True)
   ... del arrow_table
   ... print(df)
       start  prop.name  prop.start
   0  string        foo           0 
   1  string        bar          10

   
Flatten Nested Data with Projections
------------------------------------

One can also use projections to flatten...
```

ShaneHarvey (changes requested):
Thanks for the PR! I have one suggestion for the schemas page otherwise I'll defer to @blink1073 and @juliusgeo.

bindings/python/docs/source/schemas.rst
ShaneHarvey: Also it's worth pointing out that flatten() only flattens one level per call as pointed out in https://issues.apache.org/jira/browse/ARROW-4090 / https://github.com/apache/arrow/issues/20683.

juliusgeo:
I did a benchmark on flatten:
```
df = lambda: find_arrow_all(coll, {}).flatten()
df1=lambda: find_arrow_all(coll,{}, projection={
            "propStart": "$prop.start",
            "propName": "$prop.name",

        })
import timeit
print(timeit.timeit(df, number=10))
print(timeit.timeit(df1, number=10))
```
If you do this benchmark for 10k documents in the collection you get:
```
0.31618929199612467
0.2380284999962896
```
For 90k, it diverges even more:
```
2.3457120419989224
1.6852088330051629
```
I also made a pretty graph: 
image

In light of this, could you amend the documentation of flatten to indicate that it is not as performant as using projections for very large numbers of documents.

juliusgeo:
@sibbiii I'm realizing that the changes we are asking for are quite large considering the initial scope of this PR. Would you like to make the changes mentioned above? Otherwise, we are happy to merge this PR as-is and make the changes ourselves in a follow-up PR.

sibbiii:
@juliusgeo 

> I'm realizing that the changes we are asking for are quite large

I am using open source, so I also want to give back to the community. I am just very busy these days with other things, but hope to soon follow up with this PR. I just want to try out some things and benchmark them before I answer because it may depend on the use-case. And as I have to get things fixed for myself anyhow, its not much extra work ...

sibbiii:
Split into two changes -> will open two new PRs

PR #139, opened 2023-03-20
ARROW-155 Ensure compatibility with Pandas 2.0

blink1073:


.github/workflows/test-python.yml
```diff
@@ -85,6 +85,7 @@ jobs:
       - name: Install the test dependencies
         run: |
           python -m pip install -v -e ".[test]"
+          python -m pip install -U --pre pandas
```
juliusgeo, line 88: should we be installing with a direct call to pip rather than requirements.txt?

blink1073: Removed

bindings/python/pymongoarrow/pandas_types.py
```diff
@@ -147,7 +147,10 @@ def isna(self):
         )
 
     def __eq__(self, other):
-        return self.data == other
+        try:
+            return self.data == other
+        except Exception:
+            return self.data == np.array(other, dtype=object)
```
ShaneHarvey, line 153: Exception seems very broad here. Is there something more specific we can use?

bindings/python/pymongoarrow/pandas_types.py
```diff
@@ -147,7 +147,10 @@ def isna(self):
         )
 
     def __eq__(self, other):
-        return self.data == other
+        try:
+            return self.data == other
+        except Exception:
```
juliusgeo, line 152: Can this exception be more specific?

bindings/python/pymongoarrow/pandas_types.py
blink1073: Fixed to be more precise

bindings/python/pymongoarrow/pandas_types.py
blink1073: Fixed

bindings/python/pymongoarrow/pandas_types.py
```diff
@@ -147,6 +147,9 @@ def isna(self):
         )
 
     def __eq__(self, other):
+        # Binary types do not support element-wise comparison.
+        if isinstance(other, Binary):
+            other = np.array(other, dtype=object)
```
ShaneHarvey, line 152: I think I'm missing something here. Why would a PandasBSONExtensionArray ever compare equal to a bson Binary?

blink1073: I moved to the the subclass

ShaneHarvey: I'm still confused. Why does an array need to compare equal to a scalar?

blink1073: Normally it doesn't, but it is part of the test suite that pandas has for extension types.

PR #140, opened 2023-03-21
ARROW-146 Update benchmark tests to include BSON Arrays

juliusgeo:


bindings/python/benchmark.py
```diff
@@ -159,6 +167,28 @@ def to_arrow(use_large):
     find_arrow_all(c, {}, schema=schema)
 
 
+@bench("pymongoarrow-to-arrow-arr")
+def to_arrow_arrays(use_large):
```
blink1073, line 171: Since there are no list items in the "large" collection, I'm assuming this prints a line for the large docs variant with just a small time?

juliusgeo: I made it skip the large collection so it is more clear.

ShaneHarvey (commented):
I was curious about the results of these new benchmarks and discovered that we have two benchmark files. One is bindings/python/benchmark.py and the other is bindings/python/benchmarks/benchmarks.py. It looks like only the second one is used in github actions. Can we remove one of these files in favor of the other?

juliusgeo:
@blink1073 @ShaneHarvey this is ready for review now after the Consolidation.

bindings/python/benchmarks/benchmarks.py
```diff
@@ -36,7 +36,8 @@
 assert pymongo.has_c()
 SMALL = False
 LARGE = True
-collection_names = {LARGE: "large", SMALL: "small"}
+LIST = 2
```
ShaneHarvey, line 39: Why `= 2`?

bindings/python/benchmarks/benchmarks.py
```diff
@@ -53,16 +54,25 @@
...
+)
+small.insert_many([{**base_small} for _ in range(N_SMALL_DOCS)])
 
 large_doc_keys = [c * i for c in string.ascii_lowercase for i in range(1, 101)]
 schemas[SMALL] = Schema({"x": pyarrow.int64(), "y": pyarrow.float64()})
+schemas[LIST + SMALL] = Schema({"list": pyarrow.list_(pyarrow.float64())})
 dtypes[SMALL] = np.dtype([("x", np.int64), ("y", np.float64)])
 dtypes[LARGE] = np.dtype([(k, np.float64) for k in large_doc_keys])
 schemas[LARGE] = Schema({k: pyarrow.float64() for k in large_doc_keys})
+schemas[LIST + LARGE] = Schema(
```
ShaneHarvey, line 68: `LIST + LARGE` is `2 + True`? uh?

bindings/python/benchmarks/benchmarks.py
```diff
@@ -107,6 +117,13 @@ class ProfileRead:
     of reading MongoDB data.
     """
 
+    @staticmethod
+    def exercise_table(table):
```
ShaneHarvey, line 121: Please add a comment to explain why this is needed.

bindings/python/benchmarks/benchmarks.py
```diff
@@ -53,16 +54,25 @@
...
 dtypes[SMALL] = np.dtype([("x", np.int64), ("y", np.float64)])
 dtypes[LARGE] = np.dtype([(k, np.float64) for k in large_doc_keys])
 schemas[LARGE] = Schema({k: pyarrow.float64() for k in large_doc_keys})
+schemas[LIST + LARGE] = Schema(
+    {k + "_list": pyarrow.list_(pyarrow.float64()) for k in large_doc_keys}
+)
 large = db[collection_names[LARGE]]
 large.drop()
 # 2600 keys: 'a', 'aa', 'aaa', .., 'zz..z'
-large_doc = collections.OrderedDict([(k, math.pi) for k in large_doc_keys])
```
ShaneHarvey, line 65: > # 2600 keys: 'a', 'aa', 'aaa', .., 'zz..z'

This is weird, I think we want to test a large number of keys, not a large number of really large key names. Large keys names is probably very uncommon. Can we change this to something more practical? In person you suggested `a1, a2, ..., a2600` which sounds good to me.

ShaneHarvey (changes requested):

bindings/python/benchmarks/benchmarks.py
juliusgeo: Changed them to all be the same.

bindings/python/benchmarks/benchmarks.py
juliusgeo: Fixed.

bindings/python/benchmarks/benchmarks.py
juliusgeo: Fixed.

bindings/python/benchmarks/benchmarks.py
juliusgeo: Done.

blink1073 (approved):

bindings/python/benchmarks/benchmarks.py
```diff
@@ -107,6 +115,14 @@ class ProfileRead:
...
+    # We need this because the naive methods don't always convert nested objects.
+    @staticmethod
+    def exercise_table(table):
+        [
+            [[n for n in i.values] if isinstance(i, pyarrow.ListScalar) else i for i in column]
+            for column in table.columns
+        ]
+
     def setup(self):
         db[collection_names[CUR_SIZE]].drop()
```
blink1073, line 127: This needs a guard for whether the collection exists.

blink1073 (changes requested):

bindings/python/benchmarks/benchmarks.py
ShaneHarvey: dropping a non-existant collection is fine in MongoDB:
```
>>> client.test.doesnotexist.drop()
>>> client.test.doesnotexist.drop()
```

ShaneHarvey: Oh I see what you mean:
```
                   db[collection_names[CUR_SIZE]].drop()
               KeyError: False
```

bindings/python/benchmarks/benchmarks.py
```diff
@@ -30,113 +29,192 @@
...
+    # We need this because the naive methods don't always convert nested objects.
+    @staticmethod
+    def exercise_table(table):
+        [
+            [[n for n in i.values] if isinstance(i, pyarrow.ListScalar) else i for i in column]
+            for column in table.columns
+        ]
+
+    def time_to_numpy(self):
+        pass
```
blink1073, line 135: TODO?

juliusgeo: Nested arrays are not supported in NumPy or Pandas with our current implementation IIRC.

blink1073: Gotcha, can you please add a comment then?

ShaneHarvey (approved):

ShaneHarvey (changes requested):
Sorry, I thought the code looked good but I notice the benchmark is still failling.

bindings/python/benchmarks/benchmarks.py
juliusgeo: Should be resolved now because we aren't using keys.

bindings/python/benchmarks/benchmarks.py
juliusgeo: Done.

blink1073 (approved):

blink1073:
58 minutes is probably too long, perhaps cut the number of large documents in half?

PR #142, opened 2023-03-28
Make the benchmark results more readable

ShaneHarvey:
The old pip install -vvv output spits out 10000 lines of output which makes it hard to view the actual benchmark results.


blink1073 (approved):

juliusgeo (approved):
LGTM!

PR #143, opened 2023-03-28
Migrate benchmark tests to ubuntu-22.04

ShaneHarvey:
The benchmark runs much faster on ubuntu-22.04 (~5min) vs macOS (~10-15min).


blink1073 (approved):
Thanks!

PR #145, opened 2023-04-05
ARROW-144 Support BSON Code Type

blink1073:


blink1073:
WIP: still having trouble with `Regex` implementation, and I need to add support for `MinKey` and `MaxKey`.

blink1073:
I changed this to support just the `Code` type, we can defer the other types for now.

bindings/python/pymongoarrow/pandas_types.py
```diff
@@ -282,3 +282,33 @@ def __arrow_array__(self, type=None):
...
+
+    type = Code
+
+    @classmethod
+    def construct_array_type(cls) -> Type["PandasCodeArray"]:
+        return PandasCodeArray
+
+
+class PandasCodeArray(PandasBSONExtensionArray):
+    """A pandas extension type for BSON Code Type."""
```
juliusgeo, line 299: `BSON Code data type` for consistency?

bindings/python/pymongoarrow/types.py
```diff
@@ -128,6 +138,26 @@ def to_pandas_dtype(self):
         return PandasBinary(self.subtype)
 
 
+class CodeScalar(BSONExtensionScalar):
+    _klass = Code
```
juliusgeo, line 142: `_klass` is a bit too vague--maybe `bson_class` or `extension_constructor`?

bindings/python/test/test_arrow.py
```diff
@@ -261,6 +263,7 @@ def _create_data(self):
                 "Binary": [b"1", b"23"],
                 "ObjectId": [ObjectId().binary, ObjectId().binary],
                 "Decimal128": [Decimal128(str(i)).bid for i in range(2)],
+                "Code": [str(Code(str(i))) for i in range(2)],
```
juliusgeo, line 266: Can we make this just two strings that are valid code object strings (like for binary) rather than converting back and forth?

ShaneHarvey (approved):
LGTM! Can you rename the PR to something like “Support BSON code type” before merging?

Issue #136, opened 2023-03-02
 bson_iter_type(): precondition failed: iter->raw  on find_pandas_all()


kiritbasu:
Hi, I compiled the head branch yesterday to test and I ran into the above mentioned error.

The standard tests `python -m pytest` work fine.
I also imported a number of [sample mongo datasets](https://github.com/neelabalan/mongodb-sample-dataset/blob/main/sample_mflix/users.json) and running find_pandas_all against them work fine as well.

However when I try to test against my production database : 
`df = find_pandas_all(client.test.dashboards, {})`

I get the following error : 
`/tmp/mongo-c-driver-20230103-6603-ptzsc0/mongo-c-driver-1.23.2/src/libbson/src/bson/bson-iter.c:477 bson_iter_type(): precondition failed: iter->raw`

Couldn't figure out with python -v where exactly it failed, wondering if theres an easier way in the library to show the exact row/datatype it failed on?

blink1073:
Hi @kiritbasu, you could add a print statement before any place we call `bson_iter_type` with relevant local variables in [lib.pyx](https://github.com/mongodb-labs/mongo-arrow/blob/main/bindings/python/pymongoarrow/lib.pyx), using a [source install](https://mongo-arrow.readthedocs.io/en/latest/developer/installation.html), or if you have an example document that causes the error we can take a look.

kiritbasu:
Thanks @blink1073  looks like the error happens when its trying to do `extract_document_dtype - specifically value_t = bson_iter_type(doc_iter)`

The specific type it fails on is this : 
` charts: [
        {
          title: 'My Sessions by Region',
          chartType: 'LINE'
        },
        {
          title: 'My Sessions by Country',
          chartType: 'LINE'
        },
] `
there are several instances where I have empty `chart []` - i suspect its happening when its empty.

blink1073:
Ah, yes, that makes sense, thank you!  I created https://jira.mongodb.org/browse/ARROW-158 to track this issue.

kiritbasu:
Thanks.
BTW something to consider is an ability to see some sort of progress when getting a large document. With regular pymongo I wouldn't care because I'd have to do the chunking/pagination myself, but because this is arrow, I'm more inclined to get large documents and so being able to understand where in the process I am and how much is left would be very useful for these fetch_all type operations.

davidmathers:
Hi, I'm also getting this error.

[...]

PR #149, opened 2023-04-11
ARROW-158 Fix crash when performing auto schema detection on embedded lists and subdocuments

ShaneHarvey:
https://jira.mongodb.org/browse/ARROW-158

This fixes the bug in [#136](https://github.com/mongodb-labs/mongo-arrow/issues/136) by properly handling the case where a list or document is empty. When an empty or null list or document is encountered, we skip to examining the next document.

The TODOs are related to how we call append_null() to "backfill" previous records once we discover the schema. Our schema detection differs from pyarrow's from_pylist detection. An example of the difference is in test_auto_schema_first_embedded_list_null:
```
test/test_arrow.py::TestArrowExplicitApi::test_auto_schema_first_embedded_list_null

Expected: pyarrow.Table
a: struct<b: list<item: string>>
  child 0, b: list<item: string>
      child 0, item: string
----
a: [
  -- is_valid: all not null
  -- child 0 type: list<item: string>
[null,["str"],[]]]

Actual  : pyarrow.Table
a: struct<b: list<item: string>>
  child 0, b: list<item: string>
      child 0, item: string
----
a: [
  -- is_valid:  [false,true,true]
  -- child 0 type: list<item: string>
[[],["str"],[]]]

```
I'll open a follow up ticket for that issue.

### DIFF
--- a/bindings/python/pymongoarrow/lib.pyx
+++ b/bindings/python/pymongoarrow/lib.pyx
@@ -83,7 +83,7 @@ _field_type_map = {
 }
 
 cdef extract_field_dtype(bson_iter_t * doc_iter, bson_iter_t * child_iter, bson_type_t value_t, context):
-    """Get the appropropriate data type for a specific field"""
+    """Get the appropriate data type for a specific field"""
     cdef const uint8_t *val_buf = NULL
     cdef uint32_t val_buf_len = 0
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
LGTM!

PR #152, opened 2023-04-17
ARROW-113 Remove pandas -0b1 << 63 workaround

ShaneHarvey:


ShaneHarvey:
I see this test failure locally:
```python
_____________________________________________________________ TestNulls.test_int_handling ______________________________________________________________

self = <test.test_pandas.TestNulls testMethod=test_int_handling>

    def test_int_handling(self):
        # Default integral types
        int_schema = Schema({"_id": ObjectIdType(), "int64": int64()})
        int64_arr = [(i if (i % 2 == 0) else None) for i in range(len(self.oids))]
        self.coll.insert_many(
            [{"_id": self.oids[i], "int64": int64_arr[i]} for i in range(len(self.oids))]
        )
    
        table = self.find_fn(self.coll, {}, schema=int_schema)
    
        # Resulting datatype should be float64 according to the spec for numpy
        # and pandas.
        atype = self.pytype_tab_map[int]
        self.assertType(table["int64"], atype)
    
        # Does it contain NAs where we expect?
        self.assertTrue(np.all(np.equal(isna(int64_arr), isna(table["int64"]))))
    
        # Write
        self.coll.drop()
        table_write = self.table_from_dict({"int64": int64_arr})
    
        write(self.coll, table_write)
        res_table = self.find_fn(self.coll, {}, schema=int_schema)
    
>       self.equal_fn(res_table["int64"], table_write["int64"])

bindings/python/test/utils.py:183: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
bindings/python/test/test_pandas.py:362: in equal_fn
    pandas.testing.assert_series_equal(left, right, check_dtype=False)
pandas/_libs/testing.pyx:52: in pandas._libs.testing.assert_almost_equal
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   AssertionError: Series are different
E   
E   Series values are different (50.0 %)
E   [index]: [0, 1, 2, 3]
E   [left]:  [0, 0, 2, 0]
E   [right]: [0.0, nan, 2.0, nan]

pandas/_libs/testing.pyx:167: AssertionError
=============================================================== short test summary info ================================================================
FAILED bindings/python/test/test_pandas.py::TestNulls::test_int_handling - AssertionError: Series are different
```

ShaneHarvey:
From the pandas docs I learned that "-9223372036854775808" is np.nan interpreted as an int64:
```python

In [3]: # The value in the last row is np.nan interpreted as an integer
        df.groupby('a', dropna=True).transform('ffill')
Out[3]:
                     b
0                    2
1                    3
2 -9223372036854775808
```

https://pandas.pydata.org/docs/whatsnew/v1.5.0.html?highlight=9223372036854775808

ShaneHarvey:
This is still on odd issue since locally these tests pass but on Actions they fail on all platforms.

blink1073:
Perhaps locally you need to upgrade pandas to 2.0?

ShaneHarvey:
I suspect the issue is that calling bson_iter_as_int64 on a NaN returns -9223372036854775808 on some platforms. In libbson it's just a type cast:
```c
   case BSON_TYPE_DOUBLE:
      return (int64_t) bson_iter_double (iter);
```
https://github.com/mongodb/libbson/blob/3c067ba/src/bson/bson-iter.c#L1132

ShaneHarvey:
Confirmed, casting nan to int is undefined in C. Intel CPUs return -int max which explains why we see -9223372036854775808 on actions but 0 locally (on macos arm):
```
test/test_numpy.py::TestNulls::test_int_handling <- test/utils.py
bson_iter_as_double: 0.0
bson_iter_as_int64: 0

bson_iter_as_double: nan
bson_iter_as_int64: -9223372036854775808

bson_iter_as_double: 2.0
bson_iter_as_int64: 2

bson_iter_as_double: nan
bson_iter_as_int64: -9223372036854775808
```

To workaround this we can define BSON nan -> int casting as null or 0 explicitly.

blink1073 (approved):
Great, thanks!

PR #153, opened 2023-04-18
ARROW-167 int32 should coerce BSON types the same as int64

ShaneHarvey:
https://jira.mongodb.org/browse/ARROW-167

int32 now decodes BSON types the same as int64. Values that are too large to fit in an int32 cause an OverflowError. We could decide to treat such values as null instead. What do you think?


blink1073 (approved):
LGTM!

PR #155, opened 2023-05-01
ARROW-169 Add a benchmark with extension types

blink1073:


bindings/python/benchmarks/benchmarks.py
```diff
@@ -280,6 +281,43 @@ def setup_cache(self):
...
 
 
+class ProfileReadExtensionSmall(Read):
+    schema = Schema({"x": Decimal128Type(), "y": BinaryType(10)})
+    dtypes = np.dtype(np.dtype([("x", np.object_), ("y", np.object_)]))
+
+    def setup_cache(self):
+        coll = db.benchmark
+        coll.drop()
+        base_dict = collections.OrderedDict(
```
ShaneHarvey, line 291: Why use OrderedDict?

bindings/python/benchmarks/benchmarks.py
```diff
@@ -280,6 +281,43 @@ def setup_cache(self):
...
+        print(
+            "%d docs, %dk each with %d keys"
+            % (N_DOCS, len(BSON.encode(base_dict)) // 1024, len(base_dict))
+        )
+
+
+class ProfileReadExtensionLarge(Read):
+    large_doc_keys = [f"{i}" for i in range(LARGE_DOC_SIZE)]
+    schema = Schema({k: Decimal128Type() for k in large_doc_keys})
+    dtypes = np.dtype([(k, np.float64) for k in large_doc_keys])
```
ShaneHarvey, line 307: How does Decimal128 get converted to np.float64?

ShaneHarvey (changes requested):

bindings/python/benchmarks/benchmarks.py
blink1073: It was used in the other tests, switched to dict.

bindings/python/benchmarks/benchmarks.py
blink1073: Forgot to change it to `object_`, fixed.

ShaneHarvey (approved):

Issue #164, opened 2023-06-28
ARROW-175 Bug: nested data seems to be decoded even if not in schema


sibbiii:
Hi,

I found another issue with version 1.0.1 that is a little bit tricky, 
so I spend some time to create a minimal example:

Assume you have a collection with some fields that fit a certain schema.
Now you add a new key that is not in the schema so this new key should be ignored when aggregate_arrow_all is called:

```
    collection.insert_one(
        {'_id': 0,
         'new_key': 1000000000
         })

    collection.insert_one(
        {'_id': 1,
         'new_key': 1.0e50
         })

    schema = pymongoarrow.api.Schema(
        {'_id': pa.int32()
         })

    df = pymongoarrow.api.aggregate_arrow_all(collection, [{'$sort': {'_id': 1}}], schema=schema)
```

all works as expected. So far so good. This means that aggregate_arrow_all does not break even if new fields with contradicting data types are added to the collection. 

Now the same example with a nested new key:

```
    temp_collection._collection.insert_one(
        {'_id': 0,
         'obj': {'a': 1,
                 'new_key': 1000000000}
         })

    temp_collection._collection.insert_one(
        {'_id': 1,
         'obj': {'a': 1,
                 'new_key': 1.0e50}
         })

    schema = pymongoarrow.api.Schema(
        {'_id': pa.int32(),
         'obj': {'a': pa.int32()}
         })

    df = pymongoarrow.api.aggregate_arrow_all(collection, [{'$sort': {'_id': 1}}], schema=schema)
```

This raises the error 
```
pymongoarrow\lib.pyx:284: in pymongoarrow.lib.process_raw_bson_stream
OverflowError: Python int too large to convert to C long
pymongoarrow\lib.pyx:507: OverflowError
```

**But there is no reason for this error as the `new_key` should *not* even be decoded as it is not in the schema.** 

Ps.: I could remove `new_key` by a projection if I knew that it exists (which I do not know) or define all fields that I have within the schema with a projection (that leads to crazy long projections if importing 30+ fields). Interestingly, on root level it works flawlessly, the issue just arises with nested keys so maybe this is really a bug.

Ps.: If I can help by making more tests please come back to me, I am happy to contribute if i can.

Thanks,
Sebastian

ShaneHarvey:
Thanks @sibbiii! This appears to be a bug in `Schema._get_projection`. 
```python
>>> import pymongoarrow.api
>>> import pyarrow as pa
[...]

PR #165, opened 2023-06-29
ARROW-175 fix Incorrect projection on nested fields

NoahStapp:


bindings/python/pymongoarrow/schema.py
```diff
@@ -62,8 +63,13 @@ def _normalize_mapping(mapping):
 
     def _get_projection(self):
         projection = {"_id": False}
-        for fname, _ in self.typemap.items():
-            projection[fname] = True
+        for fname, ftype in self.typemap.items():
+            if isinstance(ftype, StructType):
```
ShaneHarvey, line 67: Nice, do we need to do the same for a fields that's a list_() of structs?

bindings/python/test/test_schema.py
```diff
@@ -70,3 +70,7 @@ def test_from_bson_units(self):
     def test_from_arrow_units(self):
         schema = Schema({"field1": int64(), "field2": timestamp("s")})
         self.assertEqual(schema.typemap, {"field1": int64(), "field2": timestamp("s")})
+
+    def test_nested_projection(self):
+        schema = Schema({"_id": int64(), "obj": {"a": int64(), "b": int64()}})
+        self.assertEqual(schema._get_projection(), {"_id": True, "obj": {"a": True, "b": True}})
```
ShaneHarvey, line 76: It would be good to add a test for the case reported in https://github.com/mongodb-labs/mongo-arrow/issues/164 too.

ShaneHarvey (changes requested):

bindings/python/pymongoarrow/schema.py
NoahStapp: Something that looks like {list: [{a: ...}, {b: ...}, {c: ...}]}?

ShaneHarvey: It'll be like `Schema({"list": list_(pyarrow.struct([field("a", int32())]))})`

bindings/python/test/test_arrow.py
```diff
@@ -640,6 +640,16 @@ def test_mixed_types_int64(self):
         with self.assertRaises(OverflowError):
             find_arrow_all(self.coll, {}, projection={"_id": 0}, schema=Schema({"a": int32()}))
 
+    def test_nested_contradicting_unused_schema(self):
+        data = [{"obj": {"a": 1, "b": 1000000000}}, {"obj": {"a": 2, "b": 1.0e50}}]
+        schema = Schema({"_id": int32(), "obj": {"a": int32()}})
```
ShaneHarvey, line 645: Let's remove the "_id" since we can ignore it for this test.

bindings/python/pymongoarrow/schema.py
```diff
@@ -62,10 +63,22 @@ def _normalize_mapping(mapping):
...
+        for fname, ftype in self.typemap.items():
+            if isinstance(ftype, ListType):
+                projection[fname] = self._get_field_projection_value(ftype.value_field.type)
+            else:
+                projection[fname] = self._get_field_projection_value(ftype)
         return projection
 
+    def _get_field_projection_value(self, ftype):
+        value = True
+        if isinstance(ftype, StructType):
```
ShaneHarvey, line 75: Do we also need to handle a list of lists?

ShaneHarvey (changes requested):

bindings/python/pymongoarrow/schema.py
NoahStapp: As discussed, lists of lists are projected the same as a list.

bindings/python/pymongoarrow/schema.py
```diff
@@ -62,10 +63,24 @@ def _normalize_mapping(mapping):
 
     def _get_projection(self):
         projection = {"_id": False}
-        for fname, _ in self.typemap.items():
-            projection[fname] = True
+        for fname, ftype in self.typemap.items():
+            if isinstance(ftype, ListType):
+                projection[fname] = self._get_field_projection_value(ftype.value_field.type)
+            else:
+                projection[fname] = self._get_field_projection_value(ftype)
```
ShaneHarvey, line 70: I think we can collapse this if/else into just a single line:
```
```

NoahStapp: Good catch, since lists are projected identically to whatever they contain.

ShaneHarvey (approved):
LGTM!

blink1073 (approved):

PR #168, opened 2023-09-07
ARROW-180 Support PyArrow 13

blink1073:


blink1073:
This problem has crept up again:

```
/usr/share/miniconda/envs/mongo_arrow_documentation/lib/python3.9/site-packages/pymongoarrow/lib.cpython-39-x86_64-linux-gnu.so: undefined symbol: _ZN5arrow6StatusC1ENS_10StatusCodeERKSs
```

It was previously addressed in https://github.com/apache/arrow/issues/18582.

I propose that on Ubuntu in CI, we build the wheels using Docker and then install from the wheel to match how the 
PyArrow wheels are produced.

NoahStapp (approved):

bindings/python/setup.py
```diff
@@ -18,11 +18,6 @@
 # Whether to create libarrow symlinks on posix systems.
 CREATE_LIBARROW_SYMLINKS = os.environ.get("MONGO_CREATE_LIBARROW_SYMLINKS", "1")
 
-# Whether to add the GLIBCXX override.
-# Arrow's manylinux{2010, 2014} binaries are built with gcc <= 4.8 which predates CXX11 ABI
-# - https://uwekorn.com/2019/09/15/how-we-build-apache-arrows-manylinux-wheels.html
-ADD_GLIBCXX_OVERRIDE = os.environ.get("MONGO_ADD_GLIBCXX_OVERRIDE", "1")
```
blink1073, line 24: This was what was causing the issue.  The pyarrow wheels are now using manylinux_2_28, which has gcc 12.

blink1073:
The benchmarks will fail until the next PR, because the newer version of asv is properly failing when a benchmark fails.  I've added the appropriate fixes in this PR.

.github/workflows/release-python.yml
```diff
@@ -84,7 +87,7 @@ jobs:
 
   make_sdist:
     name: Make SDist
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
```
NoahStapp, line 90: Building on macOS instead of Ubuntu?

bindings/python/docs/source/developer/installation.rst
```diff
@@ -12,7 +12,7 @@ Command Line Tools. Additionally, you need CMake and pkg-config::
   $ brew install cmake
   $ brew install pkg-config
 
-On Linux, you require gcc 4.8, CMake and pkg-config.
+On Linux, you require gcc 12, CMake and pkg-config.
```
NoahStapp, line 15: Minor nitpick, but we should probably change the wording to "installation requires ..."

NoahStapp (changes requested):

NoahStapp:
Looks good, just a few minor comments!

.github/workflows/release-python.yml
blink1073: This was an artifact from when Ubuntu wasn't working, but it doesn't matter either way.

bindings/python/docs/source/developer/installation.rst
blink1073: Fixed.

NoahStapp (approved):
LGTM!

Issue #169, opened 2023-09-12
aggregate_arrow_all(...) >four times slower in version 1.0.2 compared to 1.0.1 with fields objects


sibbiii:
Hi,

Thanks again for fixing the bugs in Version 1.0.2. 
Unfortunately it seems that the new version loads data approx.. >four times slower in case there are nested fields in the schema.
(without nested fields there seems to be no speed difference)

Are you aware of any issue already? 
We will post a unit test to reproduce the error here soon.

Sebastian

blink1073:
Hi @sibbiii, this is captured in https://jira.mongodb.org/browse/ARROW-179.

sibbiii:
Hi @blink1073,

Thanks for this info. The issue is that version 1.0.2 is so incredibly slow now that is unusable to load large datasets. Maybe we should mention this in the release notes (version 1.0.1 is fine) as MongoDB Arrows primary purpose is to be fast.

If we can help here please let me know,
Sebastian

blink1073:
Hi @sibbiii, we are thinking of reverting to the 1.0.1 behavior and documenting the limitation.  I just wanted to verify that the 1.0.1 behavior you described in #163 was not a blocker, but more of a desired feature (which we're tracking in ARROW-179).

blink1073:
Thanks, I filed https://jira.mongodb.org/browse/ARROW-181.

PR #173, opened 2023-10-16
ARROW-181 Revert nested document behavior and document limitation

blink1073:


NoahStapp (approved):
Looks good, where did the decision to revert come from? Are we still planning to write a custom builder to alleviate the performance concerns?

blink1073:
The decision came from discussion in #169.  We have [ARROW-179](https://jira.mongodb.org/browse/ARROW-179) tracking the full fix.

PR #181, opened 2023-12-13
ARROW-194 Convert top level docs to Markdown

blink1073:


bindings/python/README.md
```diff
@@ -0,0 +1,83 @@
...
+- Documentation: Available at
+[mongo-arrow.readthedocs.io](https://mongo-arrow.readthedocs.io/en/latest/).
+
+- Author: Prashant Mital
+
+**PyMongoArrow** is a companion library to PyMongo that contains tools
+for loading MongoDB query result sets as Apache Arrow tables, Pandas
+DataFrames or NumPy arrays.
+
+``` pycon
```
ShaneHarvey, line 17: extra space between \` and pycon here

bindings/python/README.md
```diff
@@ -0,0 +1,83 @@
+# PyMongoArrow
+
+- Info: A companion library to PyMongo that makes it easy to move data between
```
ShaneHarvey, line 3: Is there a nicer way to render this info rather than a bulleted list? The rst version was in a table.

ShaneHarvey (changes requested):

bindings/python/README.md
blink1073: GFM does not support tables with no headers.

bindings/python/README.md
blink1073: fixed

bindings/python/README.md
ShaneHarvey: How about we just remove this section?

blink1073: Done, I've opened https://jira.mongodb.org/browse/ARROW-196 to track adding badges instead.

ShaneHarvey (approved):

PR #185, opened 2024-01-18
[ARROW-197] Add support for Polars in PyMongoArrow

caseyclements:


bindings/python/pymongoarrow/api.py
```diff
@@ -306,9 +384,13 @@ def _tabular_generator(tabular):
         for i in tabular.to_batches():
             for row in i.to_pylist():
                 yield row
-    elif DataFrame is not None and isinstance(tabular, DataFrame):
+    elif pd.DataFrame is not None and isinstance(
+        tabular, pd.DataFrame
+    ):  # todo how could DataFrame be None?
```
blink1073, line 389: That was when pandas was an optional dependency

caseyclements: Safe to remove then?

blink1073: Yep!

bindings/python/benchmarks/benchmarks.py
```diff
@@ -197,6 +220,7 @@ def exercise_table(table):
         ]
 
     # All of the following tests are being skipped because NumPy/Pandas do not work with nested arrays.
+    # todo - Polars DOES have a Struct type. Try implementing tests for it.
```
blink1073, line 223: There are some TODO comments in this file, can you remove them and file a ticket instead?

bindings/python/pymongoarrow/api.py
```diff
@@ -14,12 +14,14 @@
 import warnings
 
 import numpy as np
+import pandas as pd
+import polars as pl
```
blink1073, line 18: This should be an optional import

bindings/python/pymongoarrow/api.py
```diff
@@ -306,9 +386,11 @@ def _tabular_generator(tabular):
         for i in tabular.to_batches():
             for row in i.to_pylist():
                 yield row
-    elif DataFrame is not None and isinstance(tabular, DataFrame):
+    elif isinstance(tabular, pd.DataFrame):
         for row in tabular.to_dict("records"):
             yield row
+    elif isinstance(tabular, pl.DataFrame):
```
blink1073, line 392: We'll need a similar guard here to what we had for pandas

bindings/python/pyproject.toml
```diff
@@ -38,7 +38,8 @@ dependencies = [
     "pyarrow >=15.0,<15.1",
     "pymongo >=4.4,<5",
     "pandas >=1.3.5,<3",
-    "packaging >=23.2,<24"
+    "packaging >=23.2,<24",
+    "polars >= 0.20.4,<2.0.0"
```
blink1073, line 42: This will have to be moved to test deps

blink1073 (changes requested):

bindings/python/benchmarks/benchmarks.py
caseyclements: Removed  [ARROW-217](https://jira.mongodb.org/browse/ARROW-217)

bindings/python/pymongoarrow/api.py
caseyclements: When `pandas` was optional, there were no exceptions raised if one called `find_pandas_all`.
With `polars` I think we should. Instead of `pa.Table.to_pandas` which exists and will throw a decent exception, I am using `polars.DataFrame.from_arrow` which won't be available. What type of exception should I raise?

caseyclements: Something like this one?

```
        raise ValueError(
            f"Invalid tabular data object of type {type(tabular)} \n"
            "Please ensure that it is one of the supported types: "
            "DataFrame, Table, or a dictionary containing NumPy arrays."
        )
```

caseyclements: This doesn't feel elegant to me. Suggestions?
```
    if pl is None:
        raise ValueError("polars is not imported. Try pip install polars.")
```

bindings/python/pyproject.toml
caseyclements: Done.

bindings/python/pymongoarrow/api.py
caseyclements: I've made changes and pushed.

bindings/python/pymongoarrow/api.py
caseyclements: Done.

blink1073 (approved):
LGTM!

PR #187, opened 2024-01-28
[ARROW-205] Add xunit-results to .gitignore

caseyclements:
Same DevEx improvement as done in mongo-python-driver.

### DIFF
--- a/bindings/python/.gitignore
+++ b/bindings/python/.gitignore
@@ -43,5 +43,6 @@ libbson
 mongo-c-driver-*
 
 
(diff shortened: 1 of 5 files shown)


bindings/python/.gitignore
```diff
@@ -43,5 +43,6 @@ libbson
 mongo-c-driver-*
 
 
-# Benchmark output files
+# Benchmark and test output files
```
blink1073, line 46: There are some unrelated changes in this PR (extra commits).

PR #191, opened 2024-01-30
[ARROW-210] Add support for large_list and large_string PyArrow DataTypes

caseyclements:
Adding support for two additional Arrow DataTypes: `large_list` and `large_string` appear in the `Table.schema` when one calls `Polars to_arrow`.

This is a small extension of our ListBuilder and StringBuilder classes and additional tests,

### DIFF
--- a/bindings/python/pymongoarrow/schema.py
+++ b/bindings/python/pymongoarrow/schema.py
@@ -88,13 +88,19 @@ class Schema:
         return False
 
     @classmethod
-    def from_arrow(cls, aschema: pa.schema):
+    def from_arrow(cls, aschema: pa.Schema):
+        """Create a :class:`~pymongoarrow.schema.Schema` instance from a :class:`~pyarrow.Schema`
+
+        :Parameters:
+          - `aschema`: PyArrow Schema
+        """
         self = cls({})
         for field in aschema:
             self.typemap[field.name] = field.type
         return self
 
     def to_arrow(self):
+        """Output the Schema as an instance of class:`~pyarrow.Schema`."""
         fields = []
         for name, type_ in self.typemap.items():
             fields.append(pa.field(name=name, type=type_))
--- a/bindings/python/pymongoarrow/schema.py
+++ b/bindings/python/pymongoarrow/schema.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 import collections.abc as abc
 
-from pyarrow import ListType, StructType
+import pyarrow as pa
 
 from pymongoarrow.types import _normalize_typeid
 
@@ -73,9 +73,9 @@ class Schema:
 
     def _get_field_projection_value(self, ftype):
         value = True
-        if isinstance(ftype, ListType):
+        if isinstance(ftype, pa.ListType):
             return self._get_field_projection_value(ftype.value_field.type)
-        if isinstance(ftype, StructType):
+        if isinstance(ftype, pa.StructType):
             projection = {}
             for nested_ftype in ftype:
                 projection[nested_ftype.name] = True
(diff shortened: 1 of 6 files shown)


bindings/python/pymongoarrow/types.py
```diff
@@ -266,6 +266,9 @@ def get_numpy_type(type):
     _atypes.is_boolean: _BsonArrowTypes.bool,
     _atypes.is_struct: _BsonArrowTypes.document,
     _atypes.is_list: _BsonArrowTypes.array,
+    _atypes.is_large_binary: _BsonArrowTypes.binary,
```
blink1073, line 269: Should we have a specific test for this type?

bindings/python/test/test_datetime.py
```diff
@@ -98,38 +98,68 @@ def test_timezone_specified_in_schema(self):
...
-            {"_id": [1, 2], "data": self.expected_times},
-            ArrowSchema([("_id", int32()), ("data", timestamp("ms", tz=tz))]),
+        """Test behavior of setting tzinfo CodecOptions in Collection.with_options.
+
+        When provided, timestamp type specifiers in the schema to inherit the specified timezone.
+        Read values will maintain this information for timestamps whether schema is passed or not.
+
+        Note, this does not apply to datetimes.
+        We also test here that if one asks for a different timezone upon reading,
+        on returns the requested timezone. # TODO Confirm whether time is adjusted!!!
```
blink1073, line 108: This was confirmed, right?

blink1073 (changes requested):

bindings/python/pymongoarrow/types.py
caseyclements: I planned to leave the binary type alone. I'm glad you pointed out that line. It's the only place that it appears. if you'd like to add it, let's discuss. Otherwise, I'll delete this line.

blink1073: I'd say remove it if we're not going to test it.

caseyclements: Done.

bindings/python/test/test_datetime.py
caseyclements: Done

blink1073 (approved):
LGTM!

bindings/python/test/test_datetime.py
```diff
@@ -98,38 +98,68 @@ def test_timezone_specified_in_schema(self):
...
 
     def test_timezone_specified_in_codec_options(self):
-        # 1. When specified, CodecOptions.tzinfo will modify timestamp
-        #    type specifiers in the schema to inherit the specified timezone
-        tz = pytz.timezone("US/Pacific")
-        codec_options = CodecOptions(tz_aware=True, tzinfo=tz)
-        expected = Table.from_pydict(
-            {"_id": [1, 2], "data": self.expected_times},
-            ArrowSchema([("_id", int32()), ("data", timestamp("ms", tz=tz))]),
+        """Test behavior of setting tzinfo CodecOptions in Collection.with_options.
```
ShaneHarvey, line 101: What's the context for these datetime test changes?

caseyclements: I expanded the test.  The original wasn't testing what it appeared to. Subleties between timestamps and datetimes.

bindings/python/pymongoarrow/schema.py
```diff
@@ -86,3 +86,22 @@ def __eq__(self, other):
         if isinstance(other, type(self)):
             return self.typemap == other.typemap
         return False
+
+    @classmethod
+    def from_arrow(cls, aschema: pa.Schema):
```
ShaneHarvey, line 91: Should we make the constructor accept `pa.Schema` rather than introduce a new api method?

caseyclements: Same same but different. This was quick. I time-boxed my work, and stopped myself short of  bigger refactoring. I would love to discuss.

ShaneHarvey: The difference is that we try to avoid adding new public methods unless there's a clear need.

blink1073: We created https://jira.mongodb.org/browse/ARROW-220 after discussing further.

PR #193, opened 2024-01-31
ARROW-211 Test macOS arm64 wheel on GHA

blink1073:


.github/workflows/release-python.yml
```diff
@@ -71,8 +82,16 @@ jobs:
...
+      - name: Build MacOS Py39 Wheels
+        if: ${{ matrix.python == 'cp39' && matrix.buildplat[0] == 'macos-11' }}
+        env:
+          MACOS_TEST_SKIP: "*arm64"
+          CIBW_BUILD: cp39-macosx_*
+          MACOSX_DEPLOYMENT_TARGET: "10.14"
+        run: python -m cibuildwheel --output-dir wheelhouse
+
       - name: Build wheels
-        if: ${{ matrix.python != 'cp38' || matrix.buildplat[0] != 'macos-11' }}
```
caseyclements, line 75: Should this be != cp39?

caseyclements (changes requested):

.github/workflows/release-python.yml
blink1073: That was the removed line, the new line covers both cp38 and cp39 since they run on macos11.

caseyclements (approved):

PR #194, opened 2024-01-31
ARROW-213 Update docs config to match PyMongo

blink1073:


caseyclements (approved):
This is great. Let's merge it. I will test it when I'm adding Polars documentation.

PR #195, opened 2024-02-01
ARROW-215 Add tox config

blink1073:


blink1073:
Benchmark will fail until we merge this because there is no `tox.ini` in the previous commit.

caseyclements (approved):

PR #205, opened 2024-04-08
Code fence syntax highlight in readme

janosh:
quick clarifying question if i may: is this library solely concerned with reading? it does not handle `numpy` type conversion to BSON on DB write?

context: it appears `numpy.bool_` fails to BSON-serialize with numpy v2 https://github.com/materialsproject/fireworks/issues/522

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Example
 ```py
 from pymongoarrow.api import Schema
 
-schema = Schema({'_id': int, 'amount': float, 'last_updated': datetime})
+schema = Schema({"_id": int, "amount": float, "last_updated": datetime})
 ```
 
 You can install PyMongoArrow on your local machine using Pip:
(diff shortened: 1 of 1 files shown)


ShaneHarvey (changes requested):
Could you fix the pre-commit failures?

> is this library solely concerned with reading? it does not handle numpy type conversion to BSON on DB write?

We support writing as well via the write() function:
https://mongo-arrow.readthedocs.io/en/stable/quickstart.html#writing-to-mongodb

janosh:
pre-commit should be green now.

thanks for the quick reply, good to know about `write`! is there some function that recursively converts a dict containing `numpy`/`pandas` types to std lib Python types?

ShaneHarvey:
Yes, you can see it here: https://github.com/mongodb-labs/mongo-arrow/blob/1.3.0/bindings/python/pymongoarrow/api.py#L417

ShaneHarvey (approved):

PR #211, opened 2024-05-15
ARROW-232 Support PyArrow 16.0

blink1073:


.github/workflows/test-python.yml
```diff
@@ -57,6 +57,9 @@ jobs:
       - name: Start MongoDB on MacOS
         if: ${{ startsWith(runner.os, 'macOS') }}
         run: |
+          brew tap mongodb/brew
```
caseyclements, line 60: mongodb/brew?

caseyclements (commented):
As we're pinning the pyarrow version so strictly, I think that we need to be very explicit about it. We need to change `docs/source/installation.rst` as it currently says 13.

.github/workflows/test-python.yml
blink1073: That's the name of our tap, I followed the MacOS install [instructions](https://www.mongodb.com/docs/manual/tutorial/install-mongodb-on-os-x/) since `mongod` is not available on the macos-14 hosts.

blink1073:
> docs/source/installation.rst as it currently says 13

That file doesn't exist anymore.  I'll add a docs changes needed to the ticket for the internal docs.

caseyclements (approved):
LGTM

Issue #219, opened 2024-06-17
objectID in nested field raises when using aggregate_polars_all


sibbiii:
Hi,

I really appreciate the support for polars, but

```
collection.insert_one({'obj': {'data_to_test': bson.ObjectId()}})
pymongoarrow.api.aggregate_polars_all(collection, [], 
                                      schema=pymongoarrow.api.Schema({'obj': {'data_to_test': bson.ObjectId}}))
```

raises with `polars.exceptions.ComputeError: cannot create series from Extension("pymongoarrow.objectid", FixedSizeBinary(12), Some(""))` as the casting of FixedSizeBinary is not done for **nested** fields. 

_Ps.: For not nested fields it works fine. Loading the arrow data frame also works fine. So this is not a showstopper, but it prohibits from using the aggregate_polars_all convenience function._

See #220 for a fix

ShaneHarvey:
Should this be closed now that #220 is merged?

lazargugleta:
@ShaneHarvey yes

PR #220, opened 2024-06-17
Support for nested ObjectIDs in polars conversion

sibbiii:
Hi,

_arrow_to_polars currently has no support to cast extension types for nested fields. 
This prohibits ObjectIDs to be read in case they are in nested fields.

I could not manage the conversion with the original code, 
but I found a way to using `arrow_table_without_extensions = arrow_table.cast(schema_without_extensions)` 
to cast the schema of the whole table in one go.

The schema_without_extensions is created recursively from the old schema.
Support for lists is still to be added, should not be that hard, maybe I try tomorrow.

I am not an expert in apache arrow. My world is Pandas and Polars.
I have wrote some unit tests locally to test the code, but I do not feel confident that I have not overlooked
something, so please review carefully. 

#219

### DIFF
--- a/bindings/python/pymongoarrow/api.py
+++ b/bindings/python/pymongoarrow/api.py
@@ -305,20 +305,25 @@ def _arrow_to_polars(arrow_table):
         raise ValueError(msg)
 
     def _cast_away_extension_type(field):
-        match field.type:
-            case pa.ExtensionType():
-                return pa.field(field.name, field.type.storage_type)
-            case pa.StructType():
-                return pa.field(field.name,
-                                pa.struct([_cast_away_extension_type(nested_field) for nested_field in field.type]))
-            case pa.ListType():
-                raise NotImplementedError
-            case _:
-                return field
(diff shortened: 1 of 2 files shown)


caseyclements:
Thank you for you submission. It looks good to me. We are waiting on Polars to support ExtensionTypes, but in the meantime, I don't see why we wouldn't add this. I cannot recall why we commented out the list and struct cases before. Please give us a few days to review.

Here is the link to the mongo-arrow task: https://jira.mongodb.org/browse/ARROW-202. It contains links to the Polars issues.

caseyclements:
Hi @sibbiii . I'm sorry for the delay. I've been very busy. Would you please add a couple tests of this new functionality?

lazargugleta:
Hey @caseyclements ,
I extended the existing test for _arrow_to_polars with lists and structs.
Let me know if you require anything else.

blink1073 (approved):
LGTM, thank you!

PR #221, opened 2024-06-24
ARROW-93 Adopt flake8-rst checks

shruti-sridhar:
This PR addresses the ticket by adding flake8-rst checks (note that doc8 checks are already present).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,13 +50,13 @@ repos:
   rev: 3.9.2
   hooks:
   - id: flake8
+    files: \.rst$
     additional_dependencies: [
         'flake8-rst==0.8.0',
         'flake8-bugbear==20.1.4',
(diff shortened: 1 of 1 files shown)


blink1073:
The failures are unrelated: "The macos-11 label has been deprecated and will no longer be available after 28 June 2024."  

@shruti-sridhar can you please open an ARROW ticket for us to update `release-python.yml` to use macos-12?  

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories
https://github.com/mongodb-labs/mongo-arrow/blob/cbbb2b70548164440597cb9b558917c0a97e8a6c/.github/workflows/release-python.yml#L41

blink1073 (approved):
LGTM!

PR #223, opened 2024-06-26
ARROW-241: Checking list-types in schema support raises an error

lazargugleta:
[JIRA TICKET](https://jira.mongodb.org/browse/ARROW-241)

### DIFF
--- a/bindings/python/pymongoarrow/types.py
+++ b/bindings/python/pymongoarrow/types.py
@@ -288,8 +288,8 @@ def _normalize_typeid(typeid, field_name):
             fields.append((sub_field_name, _normalize_typeid(sub_typeid, sub_field_name)))
         return struct(fields)
     if isinstance(typeid, list):
-        if len(typeid) > 1:
-            msg = "Please provide a struct in case of multiple types in a list"
+        if len(typeid) == 0 or len(typeid) > 1:
+            msg = "Please provide a one datatype or struct in case of multiple datatypes in a list"
             raise ValueError(msg)
         return list_(_normalize_typeid(typeid[0], "0"))
     if _is_typeid_supported(typeid):
--- a/bindings/python/pymongoarrow/types.py
+++ b/bindings/python/pymongoarrow/types.py
@@ -288,8 +288,8 @@ def _normalize_typeid(typeid, field_name):
             fields.append((sub_field_name, _normalize_typeid(sub_typeid, sub_field_name)))
         return struct(fields)
     if isinstance(typeid, list):
-        if len(typeid) == 0 or len(typeid) > 1:
-            msg = "Please provide a one datatype or struct in case of multiple datatypes in a list"
+        if len(typeid) != 1:
+            msg = f"list field in schema must contain exactly one element, not {len(typeid)}"
             raise ValueError(msg)
         return list_(_normalize_typeid(typeid[0], "0"))
     if _is_typeid_supported(typeid):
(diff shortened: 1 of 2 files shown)


Jibola:
Thanks for supplying this fix @lazargugleta!

One last thing we'd need is for you to add a test case for this function. This ensures that we will not see this error again.

Jibola (changes requested):
See above.

lazargugleta:
@Jibola 
Hey thanks for the quick reply, I added the test
Hopefully in the correct place. Let me know If there is anything else needed.

bindings/python/pymongoarrow/types.py
```diff
@@ -288,7 +288,7 @@ def _normalize_typeid(typeid, field_name):
             fields.append((sub_field_name, _normalize_typeid(sub_typeid, sub_field_name)))
         return struct(fields)
     if isinstance(typeid, list):
-        return list_(_normalize_typeid(type(typeid[0]), "0"))
+        return list_(_normalize_typeid(typeid[0], "0"))
```
ShaneHarvey, line 291: What happens if the provided list has >1 value in it? Should we error or silently ignore it?

ShaneHarvey (changes requested):

bindings/python/pymongoarrow/types.py
lazargugleta: Good point, I would say we could raise with a message that the struct should be provided.

ShaneHarvey: That sounds reasonable to me. A ValueError would be appropriate.

lazargugleta: Done

bindings/python/pymongoarrow/types.py
```diff
@@ -288,7 +288,10 @@ def _normalize_typeid(typeid, field_name):
             fields.append((sub_field_name, _normalize_typeid(sub_typeid, sub_field_name)))
         return struct(fields)
     if isinstance(typeid, list):
-        return list_(_normalize_typeid(type(typeid[0]), "0"))
+        if len(typeid) == 0 or len(typeid) > 1:
```
ShaneHarvey, line 291: Suggest `if len(typeid) != 1`

bindings/python/pymongoarrow/types.py
```diff
@@ -288,7 +288,10 @@ def _normalize_typeid(typeid, field_name):
...
+            msg = "Please provide a one datatype or struct in case of multiple datatypes in a list"
```
ShaneHarvey, line 292: Suggest: `f"list field in schema must contain exactly one element, not {len(typeid)}"`

ShaneHarvey (changes requested):

bindings/python/pymongoarrow/types.py
lazargugleta: Accept

bindings/python/pymongoarrow/types.py
lazargugleta: Accept

ShaneHarvey:
LGTM, scheduling the tests.

ShaneHarvey:
Why remove test_py_empty_list_raises?

lazargugleta:
> Why remove test_py_empty_list_raises?

It tests the same case.
Only good thing about leaving it in is that is shows intended behaviour.
Should I leave it in?

ShaneHarvey:
Yeah could you add it back? It's nice to have as an extra regression test. Also our pre-commit checks are failing. You will need to install and run:
```
pip install pre-commit
pre-commit run --all-files
```

lazargugleta:
> Yeah could you add it back? It's nice to have as an extra regression test. Also our pre-commit checks are failing. You will need to install and run:
> 
> ```
> pip install pre-commit
> pre-commit run --all-files
> ```

Thanks, test added back + reformatted code according to formating

ShaneHarvey (approved):
Thanks!

Jibola (approved):

PR #226, opened 2024-08-01
ARROW-244 Add support for PyArrow 17

blink1073:


Jibola (approved):

jannikmi:
@blink1073 is it possible to create a new release with these updates please?
I need to use pyarrow 17 and with the latest version of pymongoarrow 1.4.0 I can only get pyarrow 16 installed (cf. issue #227)
Thanks in advance.

PR #240, opened 2024-10-14
ARROW-255 Update test config and address test failures

blink1073:
- Ensure clients are closed
- Address polars test failures
- Drop Python 3.8 support


caseyclements (approved):
lgtm

PR #254, opened 2024-12-17
INTPYTHON-418 Fix handling of missing document fields

blink1073:


NoahStapp (approved):

NoahStapp (changes requested):
Is the failing benchmark tracked?

blink1073:
> Is the failing benchmark tracked?

I kicked it, if it passes I'll open a ticket about flakey benchmarks.  I don't see a reason it would get slower from this PR but you never know...

NoahStapp (approved):

PR #261, opened 2025-01-21
INTPYTHON-490 Fix segmentation fault when document doesn't match schema

### LINKED ISSUE #260: Segmentation fault when document doesn't match schema
Here is the minimal `test.py` example:

```python
from pymongo.mongo_client import MongoClient
from pymongoarrow.api import Schema, find_arrow_all

if __name__ == "__main__":
    client = MongoClient()
    client.db.test.insert_one({"x": {"y": 1}})
    find_arrow_all(client.db.test, {}, schema=Schema({"x": str}))
```

Running it causes a segfault:

```shell
$ ./venv/bin/python test.py 
Segmentation fault (core dumped)
```

Relevant package versions:
```
pyarrow==18.0.0
pymongo==4.10.1
pymongoarrow==1.6.2
```

Tested both on x86 and ARM64.

blink1073:
https://jira.mongodb.org/browse/INTPYTHON-490

Fixes #260

### DIFF
--- a/.github/workflows/release-python.yml
+++ b/.github/workflows/release-python.yml
@@ -37,7 +37,7 @@ jobs:
           - buildplat: [macos-14, macosx_*]
             python: "cp39"
         include:
-          - buildplat: [macos-12, macosx_*]
+          - buildplat: [macos-13, macosx_*]
             python: "cp39"
 
     steps:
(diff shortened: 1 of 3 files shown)


aclark4life (approved):
Addition check for `builder.type_marker` LGTM!

PR #263, opened 2025-01-22
INTPYTHON-493 Fix handling of nulls in arrays

### LINKED ISSUE #257: Incorrect null and missing value handling in nested data
Hello,

I have recently found a problem with null and missing value handling in the latest version (`>=1.6.0`). I have confirmed that this bug does not occur in prior versions (`<=1.5.2`).

Our data has nested objects, one of which has multiple optional arrays that each may or may not be there. Inside of each of those optional arrays, are objects with multiple different fields, of which I'm only selecting one. The optional nested object in this example is on `field2`.

The wrong output produced by version `>=1.6.0` is:

Screenshot 2025-01-13 151618

The correct output produced by version `<=1.5.2` is:

Screenshot 2025-01-13 151744

Notice how the first value is wrongly null on the output produced by the latest version. Also, and most alarmingly, all the values are shifted to the right by one (number of nulls). This is wrong and I can confirm that the correct output is the one produced by version `<=1.5.2`. For reference, the code used to produce this was simply:

`analysis_df = analysis_collection.aggregate_polars_all([
    {"$match": mongo_query},
    {"$project": mongo_projection}
], schema=Schema(mongo_schema))`

More interesting is that there appears to be a sort of padding behavior. This can be demonstrated by another example, where the same `field2` is now a string, but otherwise under the same optional circumstances as the previous one.

This is the output produced by version `>=1.6.0` when there is one empty record preceding a non-empty record:

Screenshot 2025-01-13 151638

However, if we add another empty record, instead of one null, it becomes two nulls and everything is shifted to the right twice:

Screenshot 2025-01-13 151648

For reference, here is the correct output produced by version `<=1.5.2`:

Screenshot 2025-01-13 151811

As previously mentioned, this is the output of the `aggregate_polars_all` function in a Polars DataFrame. However, if we try to do the same thing with pandas, it simply gives the following error in all of the above cases, in both new and old versions:

`TypeError: Cannot convert numpy.ndarray to numpy.ndarray`

Is there something that I might be missing about the correct way to read optional values? To clarify, I do not believe there is anything wrong about representing empty fields with an empty list, given in the schema I specified it should expect a list. The problem is however, the null and missing value handling and the demonstrably wrong data that is produced by that behavior. For now, this most likely means my organization will be stuck on version `1.5.2`, which is something not ideal for us.

Thanks, and let me know if I can help with any more information!

### LINKED ISSUE #258: Optional fields on pyarrow structs triggers error
Hello,

When I try to include an optional field in a `pyarrow.struct` inside of a `Schema` and call `aggregate_polars_all`, the latest version (`>=1.6.0`) throws the following error:

`ArrowInvalid: Mismatching child array lengths`

For reference, I include the code used to produce this below. `field2` is the list that contains multiple structs in which `value2` is an optional value.

```python
pa_struct = {
    "value1": pyarrow.float64(),
    "value2": pyarrow.float64()
}

mongo_projection = {
    "field1": "$field1",
    "field2": "$field2",
}

mongo_schema = {
    "field1": pyarrow.string(),
    "field2": pyarrow.list_(pyarrow.struct(pa_struct))
}

df = col.aggregate_polars_all([
    {"$match": mongo_query},
    {"$project": mongo_projection}
], schema=Schema(mongo_schema))
```

Here is how the earlier version (`<=1.5.2`) correctly and conveniently assigns `null` if the value is not there:

Screenshot 2025-01-13 171019

If I try to aggregate it to a pandas DataFrame instead using the `aggregate_pandas_all` function, it always gives me the following error, regardless of which version I use:

`TypeError: Cannot convert numpy.ndarray to numpy.ndarray`

Is this intended behavior? Is there possibly a better and intended way to achieve this in the latest version?

Let me know if I can help in providing any more information.

Thanks!

blink1073:
Fixes #257.  Fixes #258.


aclark4life (approved):
Right, so add
- `parent_names`
- `parent_type`

and check `parent_type` on iteration to store null values in the right place ? LGTM

blink1073:
The root bug was that we were using the top level document `count` to handle missing values, but that is not valid for nested documents, where missing values are dependent on how many parent documents there are.

blink1073:
Separately, we we not handling `null` values in lists properly, resulting in an empty list entry where it should have been `None` in the final Python list.

PR #279, opened 2025-02-26
INTPYTHON-228 & INTPYTHON-464 Include data types in API docs and add API docs to landing page

blink1073:


blink1073:
> Is that a performance benchmark warning for a docs build?

Yeah we have a ticket for the benchmarks being flaky.

Issue #284, opened 2025-03-14
Projection fails when reading list-of-struct data structures


zpks:
The current implementation of pymongoarrow/schemas.py assings a projection of 'True' to all fields in a struct.
This does not work for structs that contain list-of-struct datatypes; this can be fixed by reworking the projection creation to use Mongo's dot notation.

Example:
read in a data field like 
"a": {
    "b": {
      "c": [
        {
          "event": {
            "$date": "2022-01-01T00:00:00.000Z"
          },
          "value": 2,
        }
      ],
}
}

the projection to read this with a schema needs to be
{"a.b.c.event": True, "a.b.c.value": True}

The current implementation of the projection returns a projection of 
{"a": {"b": {"c": {"event": True}}}, which fails as the Mongo client reads "b" as an operator.

zpks:
Potential solution:

https://github.com/mongodb-labs/mongo-arrow/pull/285

zpks:
Fixed by https://github.com/mongodb-labs/mongo-arrow/pull/285

PR #285, opened 2025-03-14
INTPYTHON-549 Rework the projection logic to allow reading list-of-struct data structures using a specified schema

zpks:
see https://github.com/mongodb-labs/mongo-arrow/issues/284

### DIFF
--- a/bindings/python/tox.ini
+++ b/bindings/python/tox.ini
@@ -8,26 +8,26 @@ requires =
 envlist =
     # Run the unit test suite
     test,
-    ; # Build libbson
-    ; build-libbson,
-    ; # Build a dist
-    ; build-dist,
-    ; # Verify that import works without test deps
-    ; import-check,
-    ; # Run the benchmark test(s)
-    ; benchmark,
-    ; # Ensure the sphinx build has no errors or warnings.
-    ; docs,
-    ; # Serve live docs
-    ; docs-serve,
-    ; # Check links of sphinx docs
-    ; linkcheck,
-    ; # Run pre-commit on all files.
-    ; lint
-    ; # Run pre-commit on all files with manual checks.
-    ; lint-manual
-    ; # Check the sdist integrity.
-    ; manifest
+    # Build libbson
+    build-libbson,
+    # Build a dist
+    build-dist,
+    # Verify that import works without test deps
+    import-check,
+    # Run the benchmark test(s)
+    benchmark,
+    # Ensure the sphinx build has no errors or warnings.
+    docs,
+    # Serve live docs
+    docs-serve,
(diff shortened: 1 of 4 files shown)


blink1073:
Thank you @zpks!  Can you please add a test that covers your reported bug to `test_arrow.py`?

zpks:
Hello! Thank you for looking, and my apologies for creating the PR prematurely

After some wrangling I was able to test locally. 
The test for the desired feature works, it should display what's added.
I initially tried using the already-existing _create_nested_data() function with nested_elem=True, however that functionality does not work.

The change I made changes the projections from {"struct_col": {"field_1": True, "field_2": True}} to {"struct_col.field_1": True, "struct_col.field_2": True} - this works better with the mongo driver, as mongo interperts the dot as separating struct and field  - and this also works for list-of-struct structures.
This meant I also had to edit the tests for the projections in order of those to run. 
I have edited the tests for the projections - as those change, because in the previous format they errored out in the mongo read.

There was still an unrelated test failing TestExplicitPolarsApi.test_exceptions_for_unsupported_polar_types
I have not touched that code, so was unsure on how to fix those

zpks:
interestingly enough my local ruff ran fine, but the github version finds a too-long line

blink1073:
> There was still an unrelated test failing TestExplicitPolarsApi.test_exceptions_for_unsupported_polar_types
I have not touched that code, so was unsure on how to fix those

If you merge from `master` that should be fixed by #288

blink1073:
We have instructions for running the linters locally, or in this case you can copy-paste the affected lines from the diff.

https://github.com/mongodb-labs/mongo-arrow/blob/main/bindings/python/CONTRIBUTING.md#running-linters

bindings/python/tox.ini
```diff
@@ -8,26 +8,26 @@ requires =
...
-    docs-serve,
-    # Check links of sphinx docs
-    linkcheck,
-    # Run pre-commit on all files.
-    lint
-    # Run pre-commit on all files with manual checks.
-    lint-manual
-    # Check the sdist integrity.
-    manifest
+    ; # Build libbson
```
blink1073, line 11: These changes need to be reverted.  :smile:

zpks:
Okay I rebased on main (I had to sync my fork first - again, my apology for the mess) and explicityly ran the linter, that fixed the issue - I suppose I had only installed pre-commit after the too-long line was committed. Thank you for your help!

bindings/python/tox.ini
zpks: whoops

blink1073 (approved):
Thanks very much!

PR #296, opened 2025-04-22
INTPYTHON-575 Fix handling of empty embedded arrays

blink1073:


bindings/python/pymongoarrow/context.py
```diff
@@ -75,7 +75,7 @@ def _parse_builder_map(builder_map):
         elif value.type_marker == _BsonArrowTypes.array.value:
             child_name = key + "[]"
             to_remove.append(child_name)
-            child = builder_map[child_name]
+            child = builder_map.get(child_name, [])
```
caseyclements, line 78: It was just this? Cool.

caseyclements (approved):
LGTM

bindings/python/pymongoarrow/context.py
blink1073: Yep

PR #298, opened 2025-04-22
INTPYTHON-592 Convert to use just and uv

blink1073:
Benchmark is expected to fail since the previous commit doesn't have the required files and `dependency-groups`.

### DIFF
--- a/bindings/python/MANIFEST.in
+++ b/bindings/python/MANIFEST.in
@@ -9,6 +9,7 @@ include pyproject.toml
 exclude RELEASE.md
 exclude asv.conf.json
 exclude uv.lock
+exclude justfile
 
 graft pymongoarrow
 


.github/workflows/test-python.yml
```diff
@@ -105,14 +100,10 @@ jobs:
...
       - name: Build docs
-        shell: bash -l {0}
-        run: |
-          NO_EXT=1 tox -e docs
+        run: just docs
       - name: Run linkcheck
-        shell: bash -l {0}
-        run: |
-          NO_EXT=1 tox -e linkcheck
+        run: just docs-linkcheck
```
NoahStapp, line 109: Should this have `NO_EXT=1` set?

NoahStapp (changes requested):

NoahStapp (approved):

PR #310, opened 2025-05-08
INTPYTHON-609 Add support for PyArrow 20

blink1073:


.readthedocs.yaml
```diff
@@ -9,7 +9,7 @@ build:
         - asdf install uv latest
         - asdf global uv latest
     install:
-        - cd bindings/python && uv sync --no-install-project --group docs
+        - cd bindings/python && uv sync --no-install-project --frozen --group docs
```
caseyclements, line 12: WIth `frozen`.. what problems were popping up?

caseyclements (approved):
lgtm

.readthedocs.yaml
blink1073: When uv was updating the lock file it was trying to build the current project, but failed since libbson wasn't installed in the enironment.

PR #311, opened 2025-05-09
INTPYTHON-392 Add support for free-threaded Python

blink1073:
- [x] Add threading tests


bindings/python/uv.lock
blink1073, line 1: You can ignore these changes, uv just self-updated the lockfile.

aclark4life (approved):
Nice!

PR #316, opened 2025-05-22
INTPYTHON-389 - Avoid Type Mismatch Conversion to NaN

NoahStapp:


bindings/python/pymongoarrow/lib.pyx
```diff
@@ -66,30 +115,43 @@ cdef class BuilderManager:
...
             else:
                 # We only use the doc_iter for binary arrays, which are handled already.
-                self.get_builder(name, ftype, <bson_iter_t *>nullptr)
+                if self.has_schema and not allow_invalid:
+                    self.get_builder(name, ftype, <bson_iter_t *>nullptr, type_name)
+                else:
+                    self.get_builder(name, ftype, <bson_iter_t *> nullptr, None)
 
-    cdef _ArrayBuilderBase get_builder(self, cstring key, bson_type_t value_t, bson_iter_t * doc_iter):
+    cdef _ArrayBuilderBase get_builder(self, cstring key, bson_type_t value_t, bson_iter_t * doc_iter, str type_name):
```
blink1073, line 154: This is confusing to me.  I think we should keep this as a boolean named `allow_invalid`, and then have a function that takes a type marker and a `value_t` and produces the correct error.

blink1073 (changes requested):

bindings/python/test/test_bson.py
```diff
@@ -21,6 +21,7 @@
 from pymongoarrow.types import ObjectId, ObjectIdType, int64, string
 
 
+# TODO: MAKE THIS BASE CLASS SUPPORT ALLOW_INVALID
```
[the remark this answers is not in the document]
NoahStapp: Good catch, already done.

bindings/python/pymongoarrow/lib.pyx
NoahStapp: That's a much better idea. `type_marker` is already the same type as `value_t`, so we can use the same lookup method to get a user-readable name.

blink1073:
It's failing to compile now

NoahStapp:
> It's failing to compile now

Yup, caused by our arrow date types being `uint8_t` and not `bson_type_t`. Investigating a fix.

blink1073:
We can safely cast it to uint8_t, here's an example from the c driver: `static const uint8_t type = BSON_TYPE_TIMESTAMP;`

bindings/python/pymongoarrow/lib.pyx
```diff
@@ -56,6 +56,55 @@ cdef const bson_t* bson_reader_read_safe(bson_reader_t* stream_reader) except? N
         raise InvalidBSON("Could not read BSON document stream")
     return doc
 
+cdef str _get_human_readable_bson_type_t(bson_type_t type_t):
```
ShaneHarvey, line 59: `_get_human_readable_bson_type_t` is a bit of a mouthful. What about "_bson_type_name"?

ShaneHarvey (changes requested):

ShaneHarvey:
I'll defer to Steve.

blink1073 (approved):
LGTM!

Issue #286, opened 2025-03-17
Pandas as optional dependency


lazargugleta:
Hey,

Would it be possible to set pandas as the optional dependency as polars is?

I can try and do it myself If you do not have the capacity at this moment.

blink1073:
Hi @lazargugleta, we would welcome a pull request.  One thing that will be a bit tricky is that we have custom pandas types in `pandas_types.py`.

lazargugleta:
Thank you for letting me know; this seems out of scope time-wise for me currently so I will close the issue.

PR #324, opened 2025-06-20
INTPYTHON-653 & INTPYTHON-620 Make pandas optional and add free-threading support on Windows

blink1073:
Addresses #286.

### DIFF
--- a/bindings/python/test/pandas_types/test_binary.py
+++ b/bindings/python/test/pandas_types/test_binary.py
@@ -14,9 +14,13 @@
 import numpy as np
 import pytest
 from bson import Binary
-from pandas.tests.extension import base
 
 from pymongoarrow.pandas_types import PandasBinary, PandasBinaryArray
+
+try:
+    from pandas.tests.extension import base
+except ImportError:
+    pytest.skip("skipping pandas tests", allow_module_level=True)
 
 try:
     base.BaseIndexTests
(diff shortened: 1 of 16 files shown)


bindings/python/test/pandas_types/test_binary.py
```diff
@@ -14,10 +14,14 @@
...
 import pytest
 from bson import Binary
-from pandas.tests.extension import base
 
 from pymongoarrow.pandas_types import PandasBinary, PandasBinaryArray
 
+try:
+    from pandas.tests.extension import base
+except ImportError:
+    pytest.skip("skipping pandas tests", allow_module_level=True)
```
caseyclements, line 23: Could you add some thing like "To include, create environment like so: `uv sync --dev --extra test --extra test-polars --extra test-pandas`"

bindings/python/test/test_numpy.py
```diff
@@ -154,6 +154,10 @@ def test_write_error(self):
                 raise awe
 
     def test_write_schema_validation(self):
+        try:
+            import pandas as pd  # noqa: F401
+        except ImportError:
+            self.skipTest("Test requires pandas")
         arrow_schema = {k.__name__: v(True) for k, v in _TYPE_NORMALIZER_FACTORY.items()}
         schema = {k: v.to_pandas_dtype() for k, v in arrow_schema.items()}
```
caseyclements, line 162: We use pandas to go from arrow to numpy?

caseyclements (changes requested):

bindings/python/test/test_numpy.py
blink1073: Just for this test, to get the right data types

PR #334, opened 2025-07-07
PYTHON-5430 Use the zizmor action

blink1073:


.github/workflows/zizmor.yml
```diff
@@ -17,16 +17,5 @@ jobs:
...
-        run: cargo install zizmor
       - name: Run zizmor
-        run: zizmor --format sarif . > results.sarif
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      - name: Upload SARIF file
-        uses: github/codeql-action/upload-sarif@39edc492dbe16b1465b0cafca41432d857bdb31a # v3
-        with:
-          sarif_file: results.sarif
-          category: zizmor
```
Jibola, line 32: So the zizmor action also installs rust and uploads the SARIF file?

Or is that just serviced elsewhere?

caseyclements (approved):
LGTM

PR #338, opened 2025-07-25
INTPYTHON-520 Ensure all parquet data types are handled

sleepyStick:
I think this is what we want??? 
edit: docs will be completed via a docs ticket but in general the only parquet types that we don't inherently already support are on the left and what they can be converted to (that we do support) are on the right. 
- uint8 -> pa.int32
- uint16 -> pa.int32
- uint32 -> pa.int64
- uint64 -> pa.int64
- int8 -> pa.int32
- int16 -> pa.int32
- float16 -> pa.float64
- float32 -> pa.float64
- float64 -> pa.float64
- timedelta -> pa.int64

### DIFF
--- a/bindings/python/pymongoarrow/api.py
+++ b/bindings/python/pymongoarrow/api.py
@@ -35,7 +35,20 @@ from bson.raw_bson import RawBSONDocument
 from numpy import ndarray
 from pyarrow import Schema as ArrowSchema
 from pyarrow import Table, timestamp
-from pyarrow.types import is_date32, is_date64
+from pyarrow.types import (
+    is_date32,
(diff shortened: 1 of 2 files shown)


blink1073:
> 3.13t tests are failing because pandas (which I import for the test) isn't installed..is that something i need to fix here?

The tests are run with and without optional deps (polars and pandas), so they need to have a skip condition when pandas isn't installed.

blink1073:
Hmm, do we want to offer a "best effort" conversion mode where we automatically make these conversions in `write()`?

blink1073 (approved):
Very nice, thank you!

PR #346, opened 2025-08-21
INTPYTHON-497 Extend rather than replace TypeRegistry in write function

blink1073:


.github/workflows/test-python.yml
```diff
@@ -143,3 +143,33 @@ jobs:
         run: just docs
       - name: Run linkcheck
         run: just docs-linkcheck
+  test_minimum:
```
NoahStapp, line 146: Is this an unrelated change?

NoahStapp (changes requested):

PR #396, opened 2026-03-12
Prep for v1.13.0 release

caseyclements:
# Changes in Version 1.13.0 (2026/03/12)

- Add Polars Extension Type support for the BSON Binary, Code, ObjectId, and Decimal128 types.
- Add ability to use threads or multiprocesses in `find_arrow_all`, `find_pandas_all`,
  `find_numpy_all`, and `find_polars_all` with the `parallelism` parameter. For example,
  `find_arrow_all(collection, {}, parallelism="threads")` would utilize threads. Other values of
  `parallelism` are `"off"` (default) and `"processes"`.
- Fix a bug where schema inference for integer fields could produce an overflow error if the first value in the field fit within the int32 range but later values did not.
- Fix a bug where parallel batch processing or
  would fail when different batches had different inferred schemas.

### DIFF
--- a/bindings/python/CHANGELOG.md
+++ b/bindings/python/CHANGELOG.md
@@ -11,8 +11,7 @@
   `find_arrow_all(collection, {}, parallelism="threads")` would utilize threads. Other values of
   `parallelism` are `"off"` (default) and `"processes"`.
 - Fix a bug where schema inference for integer fields could produce an overflow error if the first value in the field fit within the int32 range but later values did not.
-- Fix a bug where parallel batch processing or
-  would fail when different batches had different inferred schemas.
+- Fix a bug where parallel batch processing would fail when different batches had different inferred schemas.
 
 
 # Changes in Version 1.12.0 (2026/01/26)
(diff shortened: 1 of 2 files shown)


bindings/python/CHANGELOG.md
```diff
@@ -3,14 +3,16 @@
...
-# Changes in Version 1.13.0 (2026/XX/XX)
+# Changes in Version 1.13.0 (2026/03/12)
 
 - Add Polars Extension Type support for the BSON Binary, Code, ObjectId, and Decimal128 types.
 - Add ability to use threads or multiprocesses in `find_arrow_all`, `find_pandas_all`,
   `find_numpy_all`, and `find_polars_all` with the `parallelism` parameter. For example,
   `find_arrow_all(collection, {}, parallelism="threads")` would utilize threads. Other values of
   `parallelism` are `"off"` (default) and `"processes"`.
 - Fix a bug where schema inference for integer fields could produce an overflow error if the first value in the field fit within the int32 range but later values did not.
+- Fix a bug where parallel batch processing or
```
NoahStapp: Typo, extra "or".

NoahStapp (changes requested):

NoahStapp (approved):

Issue #398, opened 2026-03-13
bug: parallel find_all on empty result fails


CoderJoshDK:
When you run a parallel version of the find_all, the `pa.concat` function fails: `ArrowInvalid: Must pass at least one table`

Instead, `results` needs to be validated that it has any data and if not, skip the concat.

CoderJoshDK:
Another inconsistent thing of note; if the schema is different (similar to #371 ) but fully incompatible, it will fail but single threaded wont. For example if one value is a boolean and another is a string. It is late so I don't have an example for you right now. I am writing this for now as a note on the off chance I forget to come back to this. 

For now, I have turned off parallel on my system. When a few (life) things slow down on my end, I might revisit here to get parallelism working for me again. (If you all don't get around to it)

Jibola:
Closing as we've provided a fix here in #405

PR #405, opened 2026-05-11
INTPYTHON-951: Parallel find_all on empty result fails

sophiayangDB:
[INTPYTHON-951](https://jira.mongodb.org/browse/INTPYTHON-951)

## Changes in this PR
Addresses bug noted in issue #398. Previously, when parallel `find_all` was used and the query returned no documents, `pa.concat_tables` is called with an empty list and results in `ArrowInvalid: Must pass at least one table`. The change updates the parallel code paths to return an empty Arrow table instead of attempting to concatenate zero tables.

## Test Plan
I added tests in `test/test_arrow.py` for empty queries across both parallelism modes, with and without an explicit schema.

### Checklist for Author
- [n/a] Did you update the changelog (if necessary)?
- [yes] Is there test coverage?
- [n/a] Is any followup work tracked in a JIRA ticket? If so, add link(s).

### Checklist for Reviewer
- [ ] Does the title of the PR reference a JIRA Ticket?
- [ ] Do you fully understand the implementation? (Would you be comfortable explaining how this code works to someone else?)
- [ ] Is all relevant documentation (README or docstring) updated?

### DIFF
--- a/bindings/python/CONTRIBUTING.md
+++ b/bindings/python/CONTRIBUTING.md
@@ -27,7 +27,6 @@ PyMongoArrow uses [Ruff](https://docs.astral.sh/ruff/) for formatting and lintin
 -   Write tests and make sure they pass (make sure you have a mongod
     running on the default port, then execute `just test` from the cmd
     line to run the test suite).
--   Add yourself to doc/contributors.rst `:)`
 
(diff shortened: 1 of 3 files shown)


caseyclements (approved):
LGTM!

